PR #13, opened 2016-03-01
fixes for packaging as a standalone app

gmaclennan:
These changes fix some things and add some sugar for packaging as a standalone app. To package as an app:

`npm run package-win`
`npm run package-mac`

@substack I'll leave you to merge/comment as you see fit.


gmaclennan:
Important change: the data dir location is moved and no longer in the same folder as the repo.

It saves to [`electron.app.getPath('userData')`](https://github.com/atom/electron/blob/master/docs/api/app.md#appgetpathname)

On a Mac, when running electron manually, this is `/Users/[username]/Library/Application Support/Electron` but when running as a packaged app it is  `/Users/[username]/Library/Application Support/ecuador-map-editor`

PR #25, opened 2016-03-16
remove timeouts for faster editing

substack:
This hack monkey patches around the setTimeout() that mainline OSM needs because of write throttling, but we don't need anything of the sort. This should speed up editing.


gmaclennan:
what a hack! but what else can we do with iD?

PR #35, opened 2016-09-16
Deploy tagged Windows/macOS builds to GitHub Releases (WIP)

noffle:
## New deployment workflow

The deployment workflow for this configuration is as follows,

1. Tag a new version (e.g. 'npm version patch')
2. Push to the Github repo (e.g. 'git push origin master')

From here, Appveyor will pick up the tag, build the release and create a
new GitHub Release with the binary asset attached.

---

## TODO
- [x] Windows
 - [x] Trigger builds via `git push`ing a tag
 - [x] build artifacts are pushed to a GitHub Release
 - [x] Windows installer is packaged + included
 - [x] Installer version pulled from `package.json`
 - [x] Make appveyor release version match `package.json`
 - [x] Node script for Windows packaging
 - [x] Thread `package.json` version through electron packaging
- [x] macOS
 - [x] Node script for macOS packaging
 - [x] Basic Travis setup for macOS
 - [x] Bundling of mac build into a DMG
 - [x] Deployment of DMG to Github Releases when tagged
- [x] Documentation
 - [x] Update README with build & deploy process

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,12 @@ platform:
 # fetch repository as zip archive
 shallow_clone: true
 
+# skip mapeo-vX.Y.Z tags; we already know they build, since they're created
+# as a result of a deployment completing
+branches:
+  except:
+    - /mapeo-v.*/
+
 artifacts:
   - path: dist/Mapeo-win32-x64
     name: mapeo-win
@@ -39,13 +45,13 @@ build_script:
 test: off
 
 deploy:
-  release: mapeo-v$(appveyor_build_version)
+  release: mapeo-v$(appveyor_build_version)  # also the name of the tag that gets created
   provider: GitHub
   auth_token:
     secure: z8zjHnYUgzuwBjJw26lgwSM2T0SdRIqMUzijaL0e0po7+8/tkBsFC/+0z7MkCRUU
-  artifact: /.*\.nupkg/            # upload all NuGet packages to release assets
+  artifact: /mapeo-win.zip/
   draft: false
   prerelease: false
   on:
-    branch: master                 # release from master branch only
+    branch: /^v/                   # don't match "mapeo-*" tags, otherwise we'll get into a tag loop
     appveyor_repo_tag: true        # deploy on tag push only
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,6 @@
 # Appveyor: for producing and deploying Windows builds
 #------------------------------------------------------
 
-version: 0.1.{build}
-
 # Test against this version of Node.js
 environment:
   nodejs_version: "6.2.1"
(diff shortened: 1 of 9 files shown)


gmaclennan:
Looking good. We should additionally push the nuget package artifacts that are created by the build step. I think the electron packager creates both a diff update and the full file. I don't completely understand these, but these are what are used for auto-update and should be published to the github release to they can be downloaded by electron in the auto-update cycle.

gmaclennan:
Let's keep this PR open and add mac builds to it along with a README update with deployment/release described.

appveyor.yml
```diff
@@ -10,6 +10,12 @@ platform:
 # fetch repository as zip archive
 shallow_clone: true
 
+# skip mapeo-vX.Y.Z tags; we already know they build, since they're created
```
gmaclennan: who/what creates these tags? Appveyor?

noffle: Yes, the `deploy.release` property decides that tag.

noffle:
Are we currently creating nuget output? I don't see anything via `grep -ri nuget` or for `nupkg`.

gmaclennan:
I can't remember how it works but there are two steps: one with [electron-packager](https://github.com/electron-userland/electron-packager) that creates the exe and all the files along with it (i.e. the folder that would go in `C:\Program Files`, and the second with [electron-winstaller](https://www.npmjs.com/package/electron-winstaller) which outputs a single .exe installer. One of these steps I think creates the nupkg files.

Also, just checking and currently the version numbers are hard-coded into the npm scripts `package-win` and `package-mac`. We need to update those from the tag / package.json version, otherwise the app itself will not know which version it is. We probably need to have a separate script for this packaging rather than the long single command as the npm script.

noffle:
> We should additionally push the nuget package artifacts that are created by the build step. I think the electron packager creates both a diff update and the full file. I don't completely understand these, but these are what are used for auto-update and should be published to the github release to they can be downloaded by electron in the auto-update cycle.

My understanding is that `electron-winstaller` needs access to previous build(s) in order to create the delta updates. Once we have the auto-update server running (ideally a static site) we can set `remoteReleases` to point at it and it should(?) work.

noffle:
Now we should be seeing the windows installer building + deploying to Github Releases, with matching `package.json` version throughout all artifacts.

gmaclennan:
Looking good! Next step: remove hard-coded electron-packager arguments - for Windows at least the version number looks like it's coming from `package.json`

noffle:
Yes! I've been trying to keep a decent devlog going on all this: http://ipfs.io/ipfs/QmVBnwiPPhtdVi1TbKrs8baUoDkdJHzLU8hyBHi7BzTr9z/day-3.html

noffle:
Windows builds and deployment looks like it's in good shape now. I'm starting to look at DMG packaging and using Travis for creating macOS deployments..

noffle:
:star2: Travis builds and deployment for macOS DMGs seems to be in working condition!

https://ipfs.io/ipfs/QmeTnEtJJkvHNeHPVqZpyYAWkFDQQ9i7bJjktvBFKK2czQ/day-6.html

gmaclennan:
LGTM!

PR #36, opened 2016-10-02
Update Travis and Appveyor tokens.

noffle:


noffle:
Disregard; I think we need repo admin tokens for this!

PR #37, opened 2016-10-03
WIP: Iceboxxify build scripts.

noffle:
This is an experiment in making the build pipeline more deterministic, cacheable, and less error prone (no shared mutable state) -- not unlike UNIX pipelines.

Building a final DMG becomes a shell command like

``` sh
$ node bin/build_id_editor.js | node bin/package_macos.js | node bin/create_dmg.js
```

For more reading on this approach, take a look at https://github.com/noffle/ice-box! This isn't strictly necessary for mapeo-desktop, but I thought you might be interested. Hacking this together was a great way to battle-test the module.


noffle:
WIP: I still need to address doing local builds + running.

noffle:
Local builds work now too, and I've rebased in all of the Travis/Appveyor and packaging fixes of recent. I just need to try running each of the Windows and macOS artifacts to make sure they work, and then this ought to be ready to merge!

@gmaclennan What are your thoughts on this approach to the build pipeline? Any concerns?

gmaclennan:
@noffle This looks like it is just building iD Editor unless I'm not understanding the code? The entry point of the app is `/app.js`, which opens a browser window with `/index.html` which then loads the patched iD from `vendor/iD/iD-patched.js` and requires `/browser/main.js` which is the main app. I don't think these build scripts are building that?

noffle:
bailing on this

PR #56, opened 2016-10-20
Add local network synchronization

noffle:
This implements local network synchronization between different machines running instances of Mapeo.

Depends on https://github.com/digidem/mapeo-desktop/pull/53

---

## Problem: Dataset Identification

One challenge for network synchronization is allowing instances of the app to discern what dataset another instance is working on, and prevent cross-contamination of the `osm-p2p-db` databases.

### Solution

The future of dataset management and cross-community contributing is not yet known, so any solution here is inherently a stop-gap. Rather than baking a dataset identifier directly into `osm-p2p-db`, I made it exist at the application level.

I did this by adding of 'metadata' key/value pair in the global app presets. This resides in `./metadata.json` relative to the repo root, and is packaged into `vendor/iD/presets.json` at build time. This object stores the key `dataset_id`, which identifies a running application's dataset. This means the running Electron app can simply evaluate `require('./vendor/iD/presets.json').metadata.dataset_id` at any time to learn what dataset the application believes it is using.

This approach should be treated as a stop-gap solution until a more desirable mechanism for identifying datasets is put into place. It necessitates a manual (though automateable!) process of configuring the application with the appropriate identifier before proliferating the application binary to a particular group.

## Problem: Network Synchronization

1. How will application instances communicate over a local network?
2. How will they find each other?
3. What protocol will they use to perform replication?

### Solution

I employed [Bonjour](https://github.com/watson/bonjour) for local network discovery, which is a protocol on top of plain multicast DNS. This module is pure javascript, so no additional software needs to be installed on user machines.

When the application starts, the Electron process starts a TCP server on a random port to act as a 'network replication service'. Incoming TCP connections will be treated as duplex streams to perform direct hyperlog replication with. A Bonjour service is broadcasted then to the local network, informing all of this service's host, port, and what `dataset_id` it is making available. From here, other applications can populate a list of machines using these Bonjour service publications and decide who to replicate with that is using a compatible dataset.

This approach is future safe: future clients wishing to share multiple or aggregate datasets can host multiple replication services and publish them all over Bonjour using legacy-friendly `dataset_id`s.

## Problem: Network Sync UI

1. How will the network sync view integrate with the existing sync UI?
2. How will the different machines to sync with be presented?

### Solution

This constituted two user interface changes:

1. A 'Network Sync' button on the main app view, adjacent to the USB sync button.

2. A new 'Network Replication' view. It is much like the USB replication view, except instead of a save/open dialog to select a file a list of machines found on the local network are listed, with accompanying 'Sync' buttons. Machines with different datasets are filtered out.

## Note

Clicking 'Sync' will perform a *uni-directional* hyperlog replication sync *to* the local machine *from* the target machine. The important thing to note is that information will be passed in *one* direction. Sync could've been made to default to being two-way, but this would seem to violate The Principle of Least Surprise: users may not expect their dataset to suddenly change from underneath them.

---

There's a lot to soak in here. :grinning:  Please ask me lots of questions! :pray:

---

mockup


gmaclennan:
Rebased on master.

gmaclennan:
This is working great! We need to revisit the UX but it works for now. I changed the flow so that after getting changes from a machine you can only click OK and start again, since we do not have a sync state for each machine. There are a bunch of todos but I will merge to master and we can use this as it is for now.

Issue #77, opened 2017-01-09
Electron logging


noffle:
Integrate (or create) an electron-friendly module to log browser and node process output to files, so that errors production can be more readily debugged.

karissa:
this is written to Mapeo/data/logs/<timestamp>

PR #83, opened 2017-01-17
Electron logging

noffle:
This adds a small logging module to mapeo-desktop. It provides the following:

1. Both: A consistent `Console`-compatible interface for use in the node and
browser processes.
2. Node: logs to a text file in the user's electron application directory with
a unqiue timestamped filename.
2. Node: writes to both stdout and the log file.
3. Node: runs a tiny websocket server for browser communication.
4. Browser: connects to the node process' websocket server. (I could've used
ipc, but this will work across many windows without configuration.)
5. Browser: writes to both the websocket server and the browser's equivalent of
stdout.
6. Browser: captures browser errors and logs them.

All that's left is for us to start using the logging module in the codebase. Use is as simple as

```js
// node
var log = require('./lib/log').Node()
log('hello')

// browser
var log = require('./lib/log').Browser()
log('world')
```

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -1,0 +1,94 @@
+var path = require('path')
+var fs = require('fs')
+var Console = require('console')
+var mkdir = require('mkdirp')
+var wsock = require('websocket-stream')
+var http = require('http')
+var through = require('through2')
+
+var nodeLogger = null
+var browserLogger = null
+
+var PORT = 17509
+
+module.exports.Node = function () {
+  if (!nodeLogger) {
+    nodeLogger = setupNodeLogger()
+  }
+  return nodeLogger
+}
+
+module.exports.Browser = function () {
+  if (!browserLogger) {
+    browserLogger = setupBrowserLogger()
+  }
+  return browserLogger
+}
+
+function setupNodeLogger () {
+  var electron = require('electron')
+
+  // create log dir
+  var logDir = path.join(electron.app.getPath('userData'), 'data', 'logs')
+  mkdir.sync(logDir)
+
+  // get electron app data dir
+  var filename = (new Date()).toISOString() + '.txt'
+  var logFilename = path.join(logDir, filename)
+
+  // fs write stream to file
+  var dest = fs.createWriteStream(logFilename)
+
+  // through stream to multiplex to stdout
+  var multi = multiplex([process.stdout, dest])
+
+  // create console instance
+  var log = new Console.Console(multi, null).log
+
+  // start ws server
+  websocketServer(function (socket) {
+    socket.on('data', function (text) {
+      multi.write(text)
+    })
+  })
+
+  return log
+}
+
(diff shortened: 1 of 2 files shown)


noffle:
Addresses #77.

package.json
```diff
@@ -65,6 +65,7 @@
     "levelup": "^1.3.1",
     "lodash": "^4.15.0",
     "minimist": "^1.2.0",
+    "mkdirp": "^0.5.1",
```
gmaclennan: remove from devDependencies

lib/log.js
```diff
@@ -0,0 +1,94 @@
...
+    socket.on('data', function (text) {
+      multi.write(text)
+    })
+  })
+
+  return log
+}
+
+function setupBrowserLogger () {
+  var socket = wsock('ws://localhost:'+PORT)
```
gmaclennan: What would happen if the socket server is not running when the browser tries to log?

lib/log.js
```diff
@@ -0,0 +1,94 @@
...
+var Console = require('console')
+var mkdir = require('mkdirp')
+var wsock = require('websocket-stream')
+var http = require('http')
+var through = require('through2')
+
+var nodeLogger = null
+var browserLogger = null
+
+var PORT = 17509
```
gmaclennan: Could be choose an available randomized port?

lib/log.js
noffle: Great question. I tried, and it indeed broke the browser process. For now, I've wrapped the socket write in a try/catch so that failure isn't quite so catastrophic. In practice, the node process always starts before the browser window is created, so this hopefully won't be an issue too often. If we notice it is, we can write a module like [reconnect-ws](https://github.com/jacobbubu/reconnect-ws) that'll do reconnects, and also some logic to buffer what failed to be written until the reconnect happens.

(I tried out `reconnect-ws`, but it has native deps that are failing to compile. There's also a non-merged PR for buffering unwritable data for future writing, but it doesn't merge cleanly anymore. Might be a bit of a project..)

lib/log.js
noffle: Agreed; this'd be ideal. However, it means you'll some kind of sideband to communicate to the browser process what port the node process chose. We could use IPC, but then we need to tell each new browser window about the port. Hardcoding, for now, seems like the least painful option.

package.json
noffle: Done.

lib/log.js
gmaclennan: I think the try/catch is good for now. Failing to capture the log is not nearly as bad as breaking the browser process, and this "should not happen" anyway.

lib/log.js
gmaclennan: Ok, leave for now. Let's leave a TODO. It's the same for the main process OSM API server anyway.

Issue #32, opened 2016-09-02
Create "zoom to data" function


aliya-ryan:
It would be really handy to be able to zoom straight to where our data is in mapeo, rather than have to search around for it zooming in and out and hunting in the Ecuadorian jungle.

Perhaps there's a way, in short term, of creating a shape around the area where the data is which is visible even when really scaled out - but currently don't seem to be able to view any features when beyond a certain zoom scale?

PR #84, opened 2017-01-18
"Zoom to Data" feature.

noffle:
Addresses #32.

Enables a "Zoom to Data" menu item under the "Visualicion" submenu, which
locates the general location of the dataset and pans + zooms the map viewport
there.

### DIFF
--- a/app.js
+++ b/app.js
@@ -215,22 +215,16 @@ function handleUncaughtExceptions () {
 
 function getGlobalDatasetCentroid (done) {
   var bbox = [[-90,90],[-180,180]]
-  var loc = null
 
   var stream = osm.queryStream(bbox)
-  stream.on('data', function (doc) {
-    if (loc) return
-    if (doc.type === 'node') {
-      loc = [Number(doc.lon), Number(doc.lat)]
-      // TODO(noffle): does this actually propagate the stream close back to
-      // the server? This isn't a request we'd like to keep running.
-      stream.unpipe()
+  var ix = setInterval(function () {
+    var doc = stream.read()
+    if (doc && doc.type === 'node') {
+      var loc = [Number(doc.lon), Number(doc.lat)]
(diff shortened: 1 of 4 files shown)


app.js
```diff
@@ -186,3 +190,27 @@ function syncToTarget (event, target) {
     server.replicateNetwork(socket, 'pull')
   }
 }
+
+function getGlobalDatasetCentroid (done) {
+  var bbox = [[-90,90],[-180,180]]
+  var loc = null
+
+  var stream = osm.queryStream(bbox)
+  stream.on('data', function (doc) {
```
gmaclennan: Rather than put this into streaming mode, we can call `stream.read()` multiple times until we get what we want. This would avoid issues with closing the stream

noffle:
Done. Although I'm still not sure whether this will leave a dangling stream open on the other end.

app.js
```diff
@@ -186,3 +190,21 @@ function syncToTarget (event, target) {
     server.replicateNetwork(socket, 'pull')
   }
 }
+
+function getGlobalDatasetCentroid (done) {
+  var bbox = [[-90,90],[-180,180]]
+
+  var stream = osm.queryStream(bbox)
+  var ix = setInterval(function () {
```
gmaclennan: Why setInterval here? if the first `n` records are not nodes this will take `n/2 seconds` to complete.

noffle: Is there another way? Doing this in a synchronous loop will block the node process.

gmaclennan: You are right. I think my advice might have been bad. Perhaps the way to do this is with through2 and `stream.unpipe()` when we get the lat/lon? I think your concern is valid - when you added the `on('data')` listener it would have been put in streaming mode and would have continued to take processor cycles even after stopping reading.

noffle:
Unpiping logic is in! Merging!

Issue #22, opened 2016-03-15
Gracefully handle errors in main process


gmaclennan:
Right now an uncaught exception in the main process throws an unintelligible (to the user) error message, and quits the app, but leaves a process running in the background.

We need to handle this error, provide a custom error message, and gracefully quit the app.

This has been a problem when a bug pops up, we fix it, but a background electron process is still running (on windows) which maintains a lock on the osm-p2p-db, and needs to be quit manually.

Here are some guidelines about how to handle uncaught exceptions in Electron:

https://github.com/atom/electron/issues/2479

karissa:
this should be fixed

PR #86, opened 2017-01-19
Catch node process exceptions and exit gracefully.

noffle:
This reports the error in an Electron message dialog, and then
terminates the process with a non-zero exit status.

Addresses #22.

### DIFF
--- a/app.js
+++ b/app.js
@@ -17,6 +17,9 @@ if (require('electron-squirrel-startup')) return
 
 var APP_NAME = app.getName()
 
+// Set up global node exception handler
+handleUncaughtExceptions()
+
 // Path to `userData`, operating system specific, see
 // https://github.com/atom/electron/blob/master/docs/api/app.md#appgetpathname
 var userDataPath = app.getPath('userData')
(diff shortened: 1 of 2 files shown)


noffle:
See lingering TODO re internationalization of the text. cc @gmaclennan

app.js
```diff
@@ -186,3 +189,22 @@ function syncToTarget (event, target) {
     server.replicateNetwork(socket, 'pull')
   }
 }
+
+function handleUncaughtExceptions () {
+  process.on('uncaughtException', function (error) {
+    // TODO(noffle): log to logging mechanism once
+    // https://github.com/digidem/mapeo-desktop/issues/77 is in.
+    console.log('uncaughtException', error)
```
gmaclennan: Add #77 before merging.

app.js
```diff
@@ -186,3 +189,22 @@ function syncToTarget (event, target) {
...
+
+    // Show a vaguely informative dialog.
+    var opts = {
+      type: 'error',
+      buttons: [ 'OK' ],  // TODO: internationalize
```
gmaclennan: We get away with it for now, "OK" is the same in Spanish, and "Fatal Error" is close enough to "Error Fatal" for people to understand...

PR #89, opened 2017-02-09
GeoJSON export polygon detection + preset injection

gmaclennan:
- Uses iD’s polygon detection algorithm based on custom presets.
- Injects preset icons and ids into exported geojson
- Replaces ‘:’ with ‘_’ for legacy data/tag names

### DIFF
--- a/lib/polygon-feature.js
+++ b/lib/polygon-feature.js
@@ -39,6 +39,7 @@ function polygonFeature (presets) {
   })
 
   return function isPolygon (coords, tags) {
+    tags = tags || {}
     if (tags.area === 'yes') return true
     if (!isClosed(coords) || tags.area === 'no') return false
     for (var key in tags) {
--- a/lib/polygon-feature.js
+++ b/lib/polygon-feature.js
@@ -2,6 +2,13 @@ var forEach = require('lodash/forEach')
 
 module.exports = polygonFeature
 
+/**
+ * Takes a hash of preset definition and returns an `isPolygon` function
+ * that will take a geoJSON feature object as an argument and return a boolean
+ * indicating if the feature is a polygon
+ * @param {Object} presets iD Presets file see https://github.com/openstreetmap/iD/blob/master/data/presets/README.md
+ * @return {Function} isPolygon function that will return boolean whether polygon or not
+ */
 function polygonFeature (presets) {
   var areaKeys = {}
   var ignore = ['barrier', 'highway', 'footway', 'railway', 'type']
--- a/lib/polygon-feature.js
+++ b/lib/polygon-feature.js
@@ -1,0 +1,52 @@
+var forEach = require('lodash/forEach')
+
+module.exports = polygonFeature
+
+function polygonFeature (presets) {
+  var areaKeys = {}
+  var ignore = ['barrier', 'highway', 'footway', 'railway', 'type']
+
+  // whitelist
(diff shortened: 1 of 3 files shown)


lib/preset-matcher.js
```diff
@@ -0,0 +1,82 @@
...
+      for (var k in preset.tags) {
+        k = k.replace(':', '_')
+        ;(g[k] = g[k] || []).push(preset)
+      }
+    }
+  }
+
+  return match
+
+  function match (feature) {
```
noffle: Nit: if you made `index` a parameter, you could break this function out of the closure and make it pure.

lib/preset-matcher.js
```diff
@@ -0,0 +1,82 @@
...
+          match = keyMatches[i]
+        }
+      }
+    }
+
+    return match || null
+  }
+}
+
+function matchScore (preset, props) {
```
noffle: Comment idea: what does 'score' mean in this context?

lib/preset-matcher.js
```diff
@@ -0,0 +1,82 @@
+module.exports = function presetMatcher (presets) {
```
noffle: Bonus points for a comment describing what this inner module does.

server.js
```diff
@@ -4,7 +4,7 @@ var sneakernet = require('hyperlog-sneakernet-replicator')
 var net = require('net')
 
 var body = require('body/any')
-var qs = require('querystring')
+var parseUrl = require('url').parse
```
noffle: This was the only use site of `querystring`; you can drop it from `package.json`.

server.js
```diff
@@ -16,10 +16,29 @@ var randombytes = require('randombytes')
...
 var metadata = userConfig.getSettings('metadata')
+var presets = userConfig.getSettings('presets')
+
+var matchPreset = require('./lib/preset-matcher')(presets.presets)
+var isPolygonFeature = require('./lib/polygon-feature')(presets.presets)
 
 var Bonjour = require('bonjour')
 var HTTP_PORT = 3198
 
+var featureMap = function (f) {
```
noffle: Nit: switch to hoisting-style function decl and move to bottom of file.

server.js
```diff
@@ -31,6 +50,11 @@ module.exports = function (osm) {
 
   var server = http.createServer(function (req, res) {
     console.log(req.method, req.url)
+    var params = parseUrl(req.url, true).query
+    var bbox = [params.minlon, params.minlat, params.maxlon, params.maxlat]
```
noffle: Did you mean to define `bbox` inside of the handler for the export route?

lib/polygon-feature.js
```diff
@@ -0,0 +1,52 @@
...
+
+    var value = d.tags[key]
+    if (d.geometry.indexOf('area') === -1 &&
+                d.geometry.indexOf('line') !== -1 &&
+                key in areaKeys && value !== '*') {
+      areaKeys[key][value] = true
+    }
+  })
+
+  return function isPolygon (coords, tags) {
```
gmaclennan: This is cut & pasted from https://github.com/openstreetmap/iD/blob/9b7d79a3fec6f0eb5e292c3b89a2912fd8751197/modules/presets/index.js#L62-L104

gmaclennan:
LGTM

PR #93, opened 2017-02-22
Add geojson export menu item.

noffle:
A maximally minimal GeoJSON export menu interface!

cc @gmaclennan


lib/menu.js
```diff
@@ -26,6 +28,32 @@ module.exports = function (app) {
...
+            dialog.showSaveDialog({
+              title: 'Guardar como..',
+              defaultPath: 'export.geojson',
+              filters: [{name: 'GeoJSON', extensions: ['geojson']}]
+            }, function (filename) {
+              if (!filename) return
+              var bbox = [-Infinity, -Infinity, Infinity, Infinity]
+              var out = fs.createWriteStream(filename)
+              var ex = exportGeoJson(app.osm, {bbox: bbox})
+              pump(ex, out, function (err) {
```
gmaclennan: I know I said I did not need to review this, but I have just realized that here we have a period when something is happening but the state is unclear. Export in my experience can take several seconds, whereas here it would appear to the user that the information is saved and the "exportación completa!" dialog would come as a bit of a surprise. We should show a dialog with some kind of progress, if it's not too hard. If this gets complicated, let's just stick with this for now - it's better having something working sooner than getting the UX perfect.

app.js
```diff
@@ -72,6 +72,7 @@ log('preparing osm indexes..')
 osm.ready(function () {
   log('osm indexes READY')
 })
+app.osm = osm
```
gmaclennan: At some point I want to re-think how we structure this and the way data flows through the app, borrowing ideas from React, and any other functional UI app. It makes me nervous as we keep threading bits of context between different modules, but it's fine for now.

lib/menu.js
```diff
@@ -26,6 +28,32 @@ module.exports = function (app) {
...
+          click: function (item, focusedWindow) {
+            dialog.showSaveDialog({
+              title: 'Guardar como..',
+              defaultPath: 'export.geojson',
+              filters: [{name: 'GeoJSON', extensions: ['geojson']}]
+            }, function (filename) {
+              if (!filename) return
+              var bbox = [-Infinity, -Infinity, Infinity, Infinity]
+              var out = fs.createWriteStream(filename)
+              var ex = exportGeoJson(app.osm, {bbox: bbox})
```
gmaclennan: We need to pass through the presets (map) and polygon features options like we do in server.js

noffle: Done. I also moved the exporting logic into its own inner module in `lib/`.

app.js
noffle: Agreed.

lib/menu.js
noffle: I went and wrote a simple progress bar view, then of course realized: an osm-p2p-db query stream's length is *unknown* at stream-time. We could show a dialog that displayed how long we've been waiting or how many objects were processed, but no real indication of how much is left.

I'll leave my progress bar work in another branch on `origin`, should we wish to pick that back up later.

PR #99, opened 2017-05-27
osm-p2p-db index version tracking & regeneration

noffle:


noffle:
1. Agreed. I'll make this change.

2. Sure! Let's do it in a separate PR rather than complicate this one.

3. Same as 2 above.

noffle:
I've addressed (1) and (3). (2) still seems like a bigger job and non-urgent, so I'm okay with leaving that as a future TODO.

PR #107, opened 2017-09-22
id fork

karissa:
Uses a new fork of id instead of monkey patches. This PR updates openstreetmap's iD to the latest version.


noffle:
@karissa very jazzed about this PR

gmaclennan:
Do we still need a build step for this? can we just require mapeo-id directly?

gmaclennan:
The new version of iD requires background imagery to have an `id` property: https://github.com/openstreetmap/iD/blob/master/API.md#background-imagery

The `imagery.json` files that we have distributed previously do not have the `id` property since I don't think this was used before, and `imagery.json` never used to be documented.

Currently this missing `id` causes frequent errors. Perhaps we should just patch in the `id` prop using the `name` prop from existing `imagery.json`? We should `console.warn` - we should fix this as we distribute updated imagery definitions, or, better, avoid them altogether by having a UI to add imagery data without config files.

gmaclennan:
I am getting an error related to presets when I try to save any edits to the map. Attached are the preset files from my `~/Library/Application Support/Electron/`.

[Archive.zip](https://github.com/digidem/mapeo-desktop/files/1820353/Archive.zip)

karissa:
To have no build step we would need to change the way that iD is being built and exported. Right now, it's self-calling anonymous function that creates iD on the window object, called as soon as the file is loaded  -- the file isn't exporting the function as expected for `npm`.

karissa:
The other issues have been fixed though, curious if you can reproduce?

karissa:
Also in order to require it directly we'd need a way to inject the css into the index.html file. Happy to do tackle this part in a separate issues

karissa:
Also in order to require it directly we'd need a way to inject the css into the index.html file. Happy to do tackle this part in a separate issues

karissa:
Also in order to require it directly we'd need a way to inject the css into the index.html file. Happy to do tackle this part in a separate issues

PR #109, opened 2017-10-17
Fix copy

karissa:
Fixes that were proposed in the google doc to make the copy on mapeo more mapeo-related.

### DIFF
--- a/browser/main.js
+++ b/browser/main.js
@@ -16,6 +16,11 @@ window.addEventListener('hashchange', function (ev) {
   localStorage.setItem('location', location.hash)
 })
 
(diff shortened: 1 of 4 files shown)


browser/main.js
```diff
@@ -34,17 +39,23 @@ id.loadLocale = function(cb) {
   if (locale && iD.data.locales.indexOf(locale) === -1) {
     locale = locale.split('-')[0]
   }
+
   if (locale && locale !== 'en' && iD.data.locales.indexOf(locale) !== -1) {
     var localePath = id.asset('locales/' + locale + '.json')
```
karissa: would this fail on windows?

PR #122, opened 2018-02-09
fix: duplicate "This is a preview" + use i18n

ralphtheninja:
There were multiple `<p>This is a preview..</p>`. Also, use i18n instead in `welcome.js`.

Before fix:

mapeo-multiple

After fix:

screenshot from 2018-02-09 01-50-16


noffle:
thank you muchly for noticing this

ralphtheninja:
@noffle I can't help it 😄

PR #125, opened 2018-03-25
incorrect lat/lon order when asking for coordinates

karissa:
Figured out that if putting it in order Latitude, Longitude it was zooming to the side of Antarctica instead of Puto Mayo


karissa:
However, it's unclear if it should be lat, lon in the display and just zoom to lon, lat underneath?

noffle:
I think it'd be good if they were consistent! lon,lat works for me

Issue #115, opened 2017-11-02
Sign with certificate on Mac


gmaclennan:
For Windows see https://github.com/digidem/mapeo-desktop/issues/118

We need to sign the download for Mac. Unsigned the app gives an error when the user tries to open it, and can only be opened by a right-click and then "Open", and the user still receives a warning.

We have a Developer certificate from Apple. We need to integrate the signing process into the build script on travis-ci. Some instructions for signing here: https://github.com/electron-userland/electron-osx-sign/wiki/1.-Getting-Started

gmaclennan:
@karissa sorry I did not get to this for the electron-builder setup. What do you need from me now to make this work.

gmaclennan:
Fixed in 4.1.0-beta.0

PR #127, opened 2018-04-09
Use electron-builder

karissa:
for #115


karissa:
@gmaclennan you have the personal identification key in your keychain on mac, can you help by exporting that and adding it to travis, along with the password? We should follow these instructions: https://www.electron.build/code-signing

There are still some tiny things left here otherwise -- need to make sure that the resulting filenames match with our current scheme.

karissa:
lets push this for post-3.0 release

PR #142, opened 2018-04-27
feature: import a tile archive in the menubar

karissa:


lib/tile-importer.js
```diff
@@ -0,0 +1,114 @@
...
+function TileImporter (userData, defaults) {
+  if (!(this instanceof TileImporter)) return new TileImporter(userData, defaults)
+  this.editing = false
+  this.defaults = Object.assign({}, TileImporter.defaults, defaults)
+  this.userData = userData
+}
+
+TileImporter.defaults = {
+  name: 'Offline Maps',
+  type: 'tms', // TODO: what is this??
```
karissa: @noffle @gmaclennan in the future i'd like to allow people to override these basic settings if they want, but for now just trying to pick smart easy defaults. any thoughts for how to be smarter on this is very welcome!

locales/es.json
```diff
@@ -6,6 +6,9 @@
   "replication-complete": "Sinconización se ha completado exitosamente. Ya debes tener la información más reciente en tu mapa. Haga un click en OK para volver al mapa",
   "replication-progress": "Sincronizando: En progreso...",
   "menu-file": "Archivo",
+  "menu-import-tiles": "Importar Offline Map Tiles...",
```
karissa: @aliya-ryan what's a better translation for this??

lib/menu.js
```diff
@@ -10,6 +10,32 @@ module.exports = function (app) {
...
+        {
+          label: i18n('menu-import-tiles'),
+          click: function (item, focusedWindow) {
+            dialog.showOpenDialog({
+              title: i18n('menu-import-tiles'),
+              filters: [{name: 'Offline Maps'}],
+              properties: ['openFile', 'openDirectory']
+            }, function (filenames) {
+              if (!filenames) return
+              app.tiles.go(filenames[0], cb)
```
karissa: would be cool to support select-multiple

lib/tile-importer.js
LukasWillin: @karissa
Just came back from checking my pull request.
What I could find out is that 'tms' is short for Tile Map Service. When assigned in the imagery.json to 'type' it wont make a difference. So I guess this is the default value for these offline tile service definitions.
Evidence:
https://wiki.openstreetmap.org/wiki/TMS

This makes sense to me because all these tiles are later served by a local ecstatic server on localhost:5005

karissa: Nice thanks

Issue #139, opened 2018-04-20
users too easily close out mapeo if you accidentally hit sync screen and want to go back


karissa:
the cancel button is forgotten and instead people hit the X and close out mapeo on accident

PR #144, opened 2018-04-27
Manually change language

### LINKED ISSUE #81: Allow manually changing of UI language
Currently Mapeo respects the operating system language for the user-interface language. It would be useful to be able to manually override this setting, because the presets we use for each project area are normally only in one language (currently Spanish), and when using these presets from a computer in en-US then strings appear as raw ids. The current work-around is to change the operating system language, but this is tedious.

We need some kind of settings pane to change the language.

karissa:
fixes #81


locales/en.json
```diff
@@ -6,6 +6,8 @@
   "replication-complete": "Syncronization has completed successfully. You now have the most recent information in your map. Click 'OK' to return to the map.",
   "replication-progress": "Syncronizing: In progress...",
   "menu-file": "File",
+  "menu-change-language": "Change language...",
+  "menu-change-language-title": "Enter the first two characters of the language (e.g., 'en' or 'es')",
```
karissa: @aliya-ryan @gmaclennan do you have a good translation for this?

karissa:
also fixes https://github.com/digidem/mapeo-desktop/issues/139

PR #147, opened 2018-05-10
Make the app React. For great justice.

karissa:


noffle:
So much code removal! :heart:

Notes:

1. The intro sequence seems to have some unescaped `$` elements and HTML throughout, e.g. 
mapeo1 I don't see this on `master`.
2. Some warnings in the console: 
mapeo2
3. The map didn't load the first time: I see a blank white screen after going through the intro. No console errors on stdout or other than the ones above. UPDATE: restarting the app got the map to load. Maybe the intro->map transition was the problem?
4. Woah the new lat/lon entry dialog *is sick*! :D
5. The top toolbar menu doesn't change its contents when the language gets changed. I don't know if this was like this before as well.

PR #154, opened 2018-06-01
Sync multi-device over mdns 

karissa:


index.js
```diff
@@ -225,6 +228,10 @@ function createMainWindow (done) {
 
     if (argv.debug) win.webContents.openDevTools()
     win.loadURL(INDEX)
+    // TODO: what's the way to see when the window is being refreshed?
+    win.on('refresh', function () {
+      server.sync.close()
```
noffle: Is the idea that when a refresh happens sync is destroyed and then re-created?

index.js
```diff
@@ -298,7 +305,7 @@ function createMainWindow (done) {
         title: i18n('open-db-dialog'),
         properties: [ 'openFile' ],
         filters: [
-          { name: 'Mapeo Data (*.' + ext + ')', extensions: [ext, 'sync'] },
+          { name: 'Mapeo Data (*.' + ext + ')', extensions: [ext, 'sync', 'zip'] },
```
noffle: Why zip too?

package.json
```diff
@@ -60,12 +60,14 @@
     "electron-squirrel-startup": "^1.0.0",
     "electron-winstaller": "^2.3.1",
     "end-of-stream": "^1.1.0",
+    "fs-blob-store": "^5.2.1",
```
noffle: Unused dep, it looks like.

src/components/IndexesBar.js
```diff
@@ -6,12 +6,11 @@ import ProgressBar from './ProgressBar'
...
     super(props)
-    var self = this
     this.state = {
       loading: false
     }
-    ipcRenderer.on('indexes-loading', self.indexesLoading.bind(self))
-    ipcRenderer.on('indexes-ready', self.indexesReady.bind(self))
+    // ipcRenderer.on('indexes-loading', self.indexesLoading.bind(self))
+    // ipcRenderer.on('indexes-ready', self.indexesReady.bind(self))
   }
```
noffle: dead code?

src/components/SyncView.js
```diff
@@ -26,50 +27,69 @@ var messages = {
...
+    })
+  }
+
+  replicate (target) {
     var self = this
-    replicate.start(filename, function (err, res, body) {
+    if (!target) return
+    this.destroyed = false
+    this.stream = replicate.start(target)
+    this.stream.on('data', function (data) {
```
noffle: Be wary here! Setting up a `data` listener uncorks the stream and data begins flowing immediately. Depending on how the replication protocol is implemented, data could be lost before `this.stream` ever gets piped into something.

src/components/SyncView.js
```diff
@@ -26,50 +27,69 @@ var messages = {
...
   componentWillUnmount () {
     this.destroyed = true
     this.stream.destroy()
+    clearInterval(this.interval)
+    ipcRenderer.removeListener('select-file', this.selectFile.bind(this))
+  }
+
+  selectFile (event, filename) {
+    if (!filename) return
+    this.replicate({filename})
```
noffle: double parens?

index.js
karissa: mapeodata format is a .zip file

src/components/SyncView.js
karissa: so better to use a through2 stream?

src/components/SyncView.js
karissa: sorry I'm not sure I understand

noffle: Oh whoops I couldn't see that you were doing {}s and not ()s. Fancy new syntax!

src/components/SyncView.js
noffle: +1 that'd work!

index.js
karissa: yeah, because it's a new sync session, it's mostly for development

package.json
karissa: ah yeah it should use safe-fs-blob-store

src/components/SyncView.js
```diff
@@ -51,20 +53,19 @@ export default class SyncView extends React.Component {
...
-    })
+      next()
+    }), done)
+
+    function done (err) {
+      if (err) console.error(err)
+    }
   }
 
   componentWillUnmount () {
```
noffle: What happens if you unmount the view down here and `this.stream === null`?

src/components/SyncView.js
```diff
@@ -51,20 +53,19 @@ export default class SyncView extends React.Component {
     if (!target) return
     this.destroyed = false
     this.stream = replicate.start(target)
-    this.stream.on('data', function (data) {
+    pump(this.stream, through.obj(function (data, enc, next) {
```
noffle: Thinking on it, I don't think this will help solve the problem: `pump` also starts the pipeline, so data events may still be lost before someone starts using `this.stream` elsewhere.

Sorry, but I think your original solution was better, but combined with [stream#pause](https://nodejs.org/api/stream.html#stream_readable_pause) after the data event listener is set up. :)

noffle: (Fun fact: if these were pull streams, it'd be easier to set up partial pipelines, since data never starts flowing until a sink is connected.)

karissa: I wonder if this is an issue, because the stream won't be used by anyone else, only by this module

src/components/SyncView.js
karissa: I think that unmounting always happens after component mounts so stream should always exist, but could make sure!

noffle: But isn't `this.stream` created on `replicate` being called? What if the view mounts and it's not called? Sorry this could be my react ignorance showing.

src/components/SyncView.js
```diff
@@ -1,75 +1,140 @@
...
   }
 
   componentWillUnmount () {
-    this.destroyed = true
-    this.stream.destroy()
+    var self = this
+    Object.keys(this.streams).map((k) => self.streams[k].destroy())
+    this.streams = {}
+    clearInterval(this.interval)
+    ipcRenderer.removeListener('select-file', this.selectFile.bind(this))
```
noffle: Will this work? I think `bind` produces a brand new function that will fail comparisons. e.g.

```js
> console.log.bind(console) == console.log.bind(console)
false
```

src/components/SyncView.js
```diff
@@ -1,75 +1,140 @@
...
-    this.destroyed = false
-    this.stream = replicate.parseMessages(function (row, next) {
-      if (self.destroyed) return
-      var status = row.topic
-      var message = messages[status] || row.message
-      self.setState({message, status})
-      return next()
+    replicate.getTargets(function (err, targets) {
+      if (err) return self.onError(err)
+      targets = JSON.parse(targets)
```
noffle: try..catch?

src/components/SyncView.js
karissa: Yeah not sure, seems to be the recommended way I've always seen listeners removed in front-end

karissa: I think if I use bind beforehand and for both adding and removing the listener, then it works! will fix! good catch

src/components/SyncView.js
karissa: nice ... catch? ;p

PR #162, opened 2018-09-19
syncfile integration

karissa:
This will work once https://github.com/digidem/mapeo-sync/pull/4 is merged and a new version of mapeo-server is published.


noffle:
You can use mapeo-sync@3.1.0 now!

PR #166, opened 2018-09-24
fix image references

karissa:
@gmaclennan can you please review?


gmaclennan:
@karissa Turning the `id` into a URL is the correct way to go I think. There is a [pending change](https://github.com/digidem/mapeo-mobile/pull/100/files#diff-f8a034fa2c7121471af9cedca9cc8f5fR46) to the observation format which does store attachments as `[{id: 'abcde1234'}]`.

As for returning an array of objects, I think this may just end up rendering extra fields for the user which are not helpful. Currently MapFilter detects image field types by detecting a URL and checking the extension to the URL path to see if it matches `.jpg, .png, .jpeg` etc. The fieldtype autodetection [can be overridden](https://github.com/digidem/react-mapfilter/blob/next/src/containers/MapFilter.js#L150-L153) by passing a `fieldTypes` prop to `<MapFilter>`, although I have not tested this with array fields. The way I'm pretty sure it would work with current code is:

```js
feature.media = {
  abc123.jpg: 'http://localhost:3000/media/originals/abc123.jpg',
  def456.jpg: 'http://localhost:3000/media/originals/def456.jpg'
}
```

I'm pretty sure if the ids have extensions this would just work. If they don't, I'm not sure if fieldTypes works for nested fields (untested) and whether it would expect a nested object, or dot-delimited props e.g. `media.abc123`

karissa:
Your example didn't work, so I tried a couple other ways:

By using `feature.properties.media`, it doesn't show up in the sidebar as an editable property -- and the *Media* tab works as expected. However, mousing over dots on the map doesn't preview the images. Is this a bug in mapfilter you think? 

Can you unpack a little bit more what you mean by your last sentence: "If they don't, I'm not sure if fieldTypes works for nested fields (untested) and whether it would expect a nested object, or dot-delimited props e.g. media.abc123" ? I look at field types and it seems to simply be a constant that tells `react-mapfilter` how to handle the properties  -- but we're already having to convert observations to features so we can just convert it to a `media` field just fine.

karissa:
Although now that I think about it, having 'media' be a property that you can filter on is actually kind of nice so you can see observations with or without media.

karissa:
OK @gmaclennan, there are no extra properties now and image preview on mouseover works. I added a note in the code on how to turn on media filtering in the sidebar if we want to do that one day.

PR #169, opened 2018-10-01
add detecting usb drives

okdistribute:
- [x] write first pass of `/datasets` route for the server (most of this code should be moved to mapeo-server)
- [ ] show available datasets in UI (@aldopuicon - do we have an idea for this already?)


noffle:
Did we decide we were going to do usb device discovery on each client, or in mapeo-sync?

karissa:
@noffle at a minimum i'd like to move the `getDatasets` function to mapeo-sync (mapeo-core?) https://github.com/digidem/mapeo-desktop/pull/169/files#diff-e6a5b42b2f7a26c840607370aed5301aR55

okdistribute:
this is still on the roadmap but likely won't happen for some time

PR #175, opened 2018-10-09
Added logo to browser window

netsgnut:
This PR proposes adding a logo to the [`BrowserWindow`](https://electronjs.org/docs/api/browser-window#new-browserwindowoptions). I am not sure about Win/Mac platform, but at least on Linux platforms there will be a nice logo showing in the taskbar.

Before:

After:

### DIFF
--- a/index.js
+++ b/index.js
@@ -217,7 +217,11 @@ function createMainWindow (done) {
 
     var INDEX = 'file://' + path.resolve(__dirname, './index.html')
     if (!win) {
-      win = new BrowserWindow({ title: APP_NAME, show: false })
+      win = new BrowserWindow({
+        title: APP_NAME,
+        show: false,
+        icon: path.resolve(__dirname, 'static', 'mapeo_256x256.png')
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -217,7 +217,11 @@ function createMainWindow (done) {
 
     var INDEX = 'file://' + path.resolve(__dirname, './index.html')
     if (!win) {
-      win = new BrowserWindow({ title: APP_NAME, show: false })
+      win = new BrowserWindow({
+        title: APP_NAME,
+        show: false,
+        icon: path.resolve(__dirname, './static/mapeo_256x256.png')
```
karissa: This is great, can you instead use `path.join` to build the path, and without slashes, like:

```
path.join(__dirname, 'static', 'mapeo_256x256.png')
```
This will work on all platforms by normalizing the path

netsgnut: Hi @karissa, great idea! Fixed at c79781d.

PR #177, opened 2018-10-18
feat(zoom-to-data): use a low-rez node density tilemap

noffle:
Creates a new version of the "osm stats" index that builds a low resolution density map of all nodes. When zooming to data, the tile with the highest density is used.


karissa:
Seems like a good approach! would be great to get this in soon

PR #181, opened 2018-10-31
Refactor dialog code and improve sync/indexes/import progress dialogs

karissa:
* Dialogs now sit under `src/components/dialogs/*`
* Remove our wrapper Modal.js class
* When indexes are regenerating, show a dialog that blocks the screen with a progress bar
* Importing files from geojson/shapefiles now show a progress bar
* Don't allow users to close the sync dialog while it's in progress
* Add sync dialog padding around the spinner:
screenshot from 2018-10-31 15-37-48


noffle:
nice! :tada:

PR #192, opened 2019-02-27
re-enable running server headless

okdistribute:


okdistribute:
out of date, will do again on new branch

PR #196, opened 2019-03-12
HACK: refresh iD editor to fix bug where points would not appear

karissa:
Bug: in some cases points would not appear in MapEditor view

Reproduce:
1. Create some points in Map Editor (or alternatively, have some points ready)
2. Open Map Filter
3. Open Map Editor again. Points have disappeared.
4. Refresh/restart Mapeo. Points appear in Map Editor again.

Doing a window refresh in the code also seems to fix it, but it has to be after a certain period of time. See the timeout added. 

I spent an hour or so trying to figure out where in iD we could do this directly, instead of hacking it on top. I wouldn't recommend merging this pull request as it is a total hack, but it is better than people thinking they've lost their data entirely. I just haven't figured out a better way quite yet.

### DIFF
--- a/src/components/MapEditor.js
+++ b/src/components/MapEditor.js
@@ -125,6 +125,8 @@ export default class MapEditor extends React.Component {
       self.updateSettings()
     })
 
+    setTimeout(() => this.refreshWindow(), 1000)
+
     window.onbeforeunload = function () { self.id.save() }
   }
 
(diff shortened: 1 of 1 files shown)


karissa:
@gmaclennan any ideas off the top of your head?

noffle:
I didn't look at your code, but does refreshing after `osm.ready()` fires work? I'm wondering if maybe it could be the indexes catching up? (Kinda seems unlikely though, with so few points)

gmaclennan:
@karissa I was able to replicate, and spent an hour trying to figure out what is going on. Didn't reach any conclusion, but did add something almost as hacky, but slightly better user experience.

karissa:
Yeah that looks more mergable than mine @gmaclennan! Nice idea.

PR #198, opened 2019-03-29
WIP: update to latest kappa-core

karissa:
#### Questions
- [x]  installStatsIndex? Gets the dataset centroid. How might this be done now?
- [?] how can we tell with kappa-core that indexes are still loading? does this matter? (used to be `osm.ready`) @noffle has this been remedied?

### TODO (maybe?)
- [x] Keep track of open sync streams and clean them up once they're done
- [-] ~Allow syncing while the sync screen is closed~ (punting)
- [ ] ~Cancel open syncs that are stalled/not working.~ (punting, currently reports an error in the ui which allows you to close the screen)
- [ ] ~MapFilter filters not able to be changed (click `change filters...` and try to change a filter)~ (will be done in a separate PR)


noffle:
> installStatsIndex? Gets the dataset centroid. How might this be done now?

We'll need a kappa view instead. You can check out what views look like (kappa-osm/lib/ dir, or our kappa workshop!), or just comment it out for now and I can write it when I get a chance :3

> how can we tell with kappa-core that indexes are still loading? does this matter? (used to be osm.ready)

same API :tada:

gmaclennan:
@noffle @karissa I think we'd agreed to drop `osm-p2p` and move to using `osm-p2p-db`. We should switch it here - does `osm-p2p` have the latest updates even? What do you think should be our deprecation strategy for that?

index.js
```diff
@@ -370,12 +373,10 @@ function handleUncaughtExceptions () {
 }
 
 function getGlobalDatasetCentroid (done) {
-  /*
-  app.osm.stats.getMapCenter(function (err, center) {
+  app.osm.core.api.stats.getMapCenter('node', function (err, center) {
```
karissa: self-note TODO: `getMapCenter('obseravtion'` for mapfilter view

gmaclennan:
Some ideas for a sync manager API without react dependency:

```js
const sync = new SyncManager()
sync.setAvailable() // starts announcing and starts listening for other peers
sync.onUpdate(peers => this.setState({peers}))
sync.startSync(peer.id)
sync.setUnavailable() // stop announcing and stop listening
sync.cancelSync(peer.id)
// peer = { id, name, status, progress: { dbComplete, dbTotal, mediaComplete, mediaTotal} }
// status = available | syncing | complete
```

Then for using in React:

```js
class MyDialog extends React.Component {
  componentDidMount() {
    this.sync = new SyncManager()
    this.sync.setAvailable()
    this.sync.onUpdate(peers => this.setState({peers}))
  }
  componentWillUnmount() {
    this.sync.setUnavailable()
    // cancel open syncs??
  }
  render() {
    return this.state.peers.sort(sortByStatus).map(peer => (
      <Target key={peer.id} {...peer} onStartClick={this.sync.startSync} onCancelClick={this.sync.cancelSync} />
    ))
    // render rest of dialog as needed...
  }
}
```

```js
class Target extends React.PureComponent {
  this.handleStartClick() {
    this.props.onStartClick(this.props.id)
  }
  this.handleCancelClick() {
    this.props.onCancelClick(this.props.id)
  }
  render() {
    const {name, status, progress} = this.props
    // render as needed based on status & progress
  }
}
```

karissa:
@gmaclennan done in https://github.com/digidem/mapeo-desktop/pull/198/commits/8e1ffd84ce0db3a82a1f0cf1aa539ef13de0d315

PR #212, opened 2019-05-13
fix(mapfilter): make sure certain props shouldn't update

### LINKED ISSUE #211: Remove created_at and timestamp fields when editing
When we load observations to MapFilter we add the `created_at` and `timestamp` fields to the properties. We need to remove those fields when saving the observation after an update: https://github.com/digidem/mapeo-desktop/blob/master/src/renderer/components/MapFilter.js#L118-L120

karissa:
fixes #211
are there any other properties that shouldn't be updated?
wondering also if this should be handled in mapeo-core instead @gmaclennan

### DIFF
--- a/src/renderer/components/MapFilter.js
+++ b/src/renderer/components/MapFilter.js
@@ -28,6 +28,11 @@ const theme = createMuiTheme({
 
 const osmServerHost = 'http://' + remote.getGlobal('osmServerHost')
 
+const DONT_UPDATE_PROPS = [
+  'created_at',
+  'timestamp'
+]
+
 const customStyleUrl = `${osmServerHost}/styles/mapfilter-style/style.json`
 const defaultStyleUrl = `${osmServerHost}/static/style.json`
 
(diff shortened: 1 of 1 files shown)


gmaclennan:
Looks good to me.

> wondering also if this should be handled in mapeo-core instead @gmaclennan

I think mapeo-core should handle "read-only" props, like created_at, and not update them. However this case is different, because in order to work around an implementation detail of MapFilter we copy the create_at and timestamp properties alongside everything under `tags`, which is different to where they are stored when we get the observation from the API. I'm thinking we should switch MapFilter to just expect the data it receives to expect data in the structure of an observation, rather than switching between two different formats. But anyway I want to break MapFilter out into just components, rather than an "app", and have visualization components and then the edit component, so we can deal with this all much more sensibly.

> are there any other properties that shouldn't be updated?

I think it would break things if we ever changed `type`. I don't think we check for that, or test what would happen if it did change. Not an urgent issue right now. `id` should not change either. Can't think of any others off the top of my head right now.

karissa:
Yeah, MapFilter requires things in geojson which is a bit annoying for our particular use case, although it does make it more flexible for the general case. So I think keeping that functionality available as an option might make it more usable for other people (or they can convert from geojson to the observations format?)

PR #214, opened 2019-05-14
Tiles asar

karissa:
## Breaking Changes

* Map tiles and aerial imagery now must be referenced from under a `styles`  directory. If you were using background imagery in 3.x, you will need to change your URL from `/Offline-Maps/...` to `/styles/Offline-Maps/...`. 

* Presets require a namespace now, which is `default`. If you don't see your presets from a previous installation, you'll need to re-import the configuration file from **File->Import configuration**

### DIFF
--- a/src/main/tile-importer.js
+++ b/src/main/tile-importer.js
@@ -1,28 +1,32 @@
+const logger = require('../log')
 const mkdirp = require('mkdirp')
 const pump = require('pump')
 const path = require('path')
 const tar = require('tar-fs')
-const mirror = require('mirror-folder')
+const asar = require('asar')
 const fs = require('fs')
+
+var log = null
 
 module.exports = TileImporter
 
 function TileImporter (userData, defaults) {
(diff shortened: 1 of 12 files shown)


src/main/tile-importer.js
```diff
@@ -74,30 +78,33 @@ TileImporter.prototype._extractTar = function (tilesPath, destPath, cb) {
...
-      return self._moveDirectory(tilesPath, dir, cb)
-    } else return cb(new Error('Tiles format not accepted'))
+    mkdirp(tilesDest, (err) => {
+      if (err) return cb(err)
+      if (path.extname(tilesPath) === '.asar') {
+        var filename = path.basename(tilesPath)
+        return fs.copyFile(tilesPath, path.join(tilesDest, filename), cb)
+      }
+      if (stat.isDirectory()) {
+        var styleId = path.basename(tilesDest)
```
karissa: we're still supporting directories in the code, even if we aren't exposing that through the user interface. In a future iteration we can imagine folks dragging and dropping instead of selecting from a menu, which would make this use case much easier

PR #226, opened 2019-06-21
Sidebar UX redesign

ChrisC:
## adds the newly re-designed sidebar
Screen Shot 2019-06-21 at 8 39 37 AM

- wasn't sure what to do with the export menu items (they seem only relevant to the MapEditor?) so they're only rendered in the MapEditor view. Let me know if they should always be rendered.

---

Screen Shot 2019-06-21 at 8 26 55 AM

- the Sync menu item still brings up the old SyncView dialog for now. Getting to that re-design next.


package.json
```diff
@@ -58,6 +58,7 @@
     "@material-ui/core": "^1.5.1",
     "@material-ui/icons": "^3.0.1",
     "JSONStream": "^1.3.5",
+    "appdmg": "^0.4.5",
```
gmaclennan:
```suggestion
```

This shouldn't be committed since it's a Mac only dep. On the todo list: figure out how to stop this getting added to package.json on `npm install`

okdistribute:
is this stale now?

PR #227, opened 2019-06-27
Sync ux redux

ChrisC:
We now have a screen for syncing instead of a modal! 
Screen Shot 2019-06-27 at 1 27 31 PM

(this branch was based off of the [sidebar redux branch](https://github.com/digidem/mapeo-desktop/pull/226) so that should be merged first)


gmaclennan:
Advancing well. Most of my comments are design niggles. Happy for @aldopuicon to offer suggestions:

- The interaction is a little strange: it looks like there is a sync button, but it doesn't interact like a button, instead the whole square acts like a button. Interested on Aldo's thoughts on this, but it doesn't feel right to me.

- The text in the sync square is centered to the square, without taking into consideration the sync "button" so it looks off-center (the designs center the text to the space between the button and top fo square).

- The sync squares have rounded corners in the designs

- The icon in the sync button is different to the icon in the bar on the left.

- The icons in the sync buttons are not aligned.

- There is an intermediate state after clicking the sync button when the button disappears, then appears with the progress indicator

- It would be good if the sync buttons were similar to the mobile ones and looked about the same.

- Typo: Dispositivos. I think maybe "Dispositivos disponibles" is maybe a better translation?

- "Last completed" is in english.

locales/es.json
```diff
@@ -65,6 +65,7 @@
   "sync-file-info": "por Archivo",
   "sync-searching-targets": "Buscando otros dispotivos",
   "sync-available-devices": "Dispotivos listos",
+  "sync-searching-tip": "Asegúrese de que los dispositivos estén encendidos y conectados a la misma red wi-fi.",
```
ScottBrenner:
```suggestion
  "sync-searching-tip": "Asegúrese de que los dispositivos estén encendidos y conectados a la misma red Wi-Fi.",
```

okdistribute:
did this get merged? seems like it's stale

Issue #219, opened 2019-05-30
.deb installer shows incorrect metadata


okdistribute:
The stuff under 'details' is so wrong!
![Screenshot from 2019-05-30 14-19-33](https://user-images.githubusercontent.com/633012/58665559-ff11f780-82e5-11e9-9e10-1994298198b0.png)

arky:
@karissa Unfortunately this bug cannot be resolved until the Ubuntu/Debian upstream issues with machine readable license file are fixed.  (See https://bugs.launchpad.net/ubuntu/+source/software-center/+bug/435183)

Recommend closing this bug as it can not be resolved in near future. :(

arky:
@okdistribute  Guess this needs to be closed as invaild/wontfix

okdistribute:
Thanks arky!

PR #229, opened 2019-07-02
Correcting the invalid SPDX "license" expression

arky:
Resolves the minor warning " warn mapeo-desktop@4.0.4-beta license should be a valid SPDX license expression" See https://spdx.org/licenses/ for more description. 

This field is also used  for generating .deb package metadata  #219


noffle:
Thank you @arky!

PR #230, opened 2019-07-03
Adding GH_TOKEN step to build process.

arky:
I think it is essential for new contributors to export a GH_TOKEN before they do 'npm run dist'.


noffle:
Thanks @arky. I think this is only needed for `dist`, not just `pack`ing. I'll further update to clarify.

Issue #237, opened 2019-07-25
Remove "--arch=x64" option from leveldb rebuild options


arky:
Perhaps removing "--arch=x64" hard limit would allow building and running integration tests on windows 32-bit machines. There are still machines running older 32-bit windows operating system in the global south.  

`    "rebuild-leveldb": "cd node_modules/leveldown && cross-env HOME=~/.electron-gyp node-gyp rebuild --target=2.0.7 --runtime=electron --arch=x64 --dist-url=https://atom.io/download/atom-shell",`

https://github.com/digidem/mapeo-desktop/blob/d52be7f82642fae072c55427871581cc609fa690/package.json#L29

PR #240, opened 2019-07-26
Remove --arch=x64 option

arky:
Removing "--arch=x64" hard limit would allow building and running integration tests on windows 32-bit machines. Without this option, the arch setting will be picked from environment.  

Fixes https://github.com/digidem/mapeo-desktop/issues/237

Tested on Windows 7 32-bit machine


gmaclennan:
Thanks @arky I don't know why we had the x64 limit before. This looks good. I tried cleaning the caches on Travis to see if that fixes build failures on travis, which do not seem related to this PR.

PR #246, opened 2019-09-04
Enclose tile-dl output paramters in double quotes

arky:
In Windows Powershell the output parameters needs to be within double quotes ( " " ). I believe otherwise the curly braces ({}) are interpreted as Unicode escape sequence


noffle:
Thanks!

PR #259, opened 2019-12-05
Background process IPC

okdistribute:
This now should make the application more speedy. All @mapeo/core related computation, including HTTP server stuff is now happening in a background process (if in production) or a background window (if in development mode).

This doesn't depend on any changes to mapeo/core or server, although while writing this code I found a bunch of ways we could improve things.. but I think it's better to look at those in later PRs.

The code in the following list could be re-used in the mobile application:

* `src/main/mapeo.js` -> most of this could be refactored into mapeo/core 
* `src/main/mapeo-ipc.js` -> main process ipc
* `src/renderer/new-api.js` -> ipc-client 

Ideally they could be refactored into other modules, happy to do that -- it'd also then make this PR much smaller. It's a tradeoff between touching many modules or sticking with one change... 

## Known issues
- ~[ ] Exporting GeoJSON without presets fails silently~ *this is a bug on master* (https://github.com/digidem/mapeo-desktop/issues/318)
- ~[ ] Exporting observations from MapFilter doesn't tell you when it's done~ *this is a bug on master too* (#319)
- [x] Errors should be displayed in the frontend (using an alert box would be feature-partiy with master branch)
- [x] Publish the middleware library @okdistribute


.babelrc
```diff
@@ -1,5 +1,6 @@
 // This file is formatted as JSON5 https://github.com/json5/json5
 {
+  sourceType: "unambiguous",
```
okdistribute: This was necessary to allow require and import in the same file

package.json
```diff
@@ -71,6 +71,7 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.24",
     "@fortawesome/free-solid-svg-icons": "^5.11.1",
     "@fortawesome/react-fontawesome": "^0.1.4",
+    "@mapeo/core": "^8.1.3",
```
okdistribute: this is for the errors, which makes me think even more that all of that code in src/main/mapeo.js ought to instead be in mapeo/core itself.

package.json
```diff
@@ -3,7 +3,7 @@
   "name": "mapeo-desktop",
   "description": "Mapping and monitoring in remote environments.",
   "productName": "Mapeo",
-  "version": "5.0.4",
+  "version": "5.0.6",
```
okdistribute: this PR also includes a merge commit from master.. nasty but there's lots of big changes so it was the best way without having rebase hell

src/middleware/background-ipc.js
```diff
@@ -0,0 +1,53 @@
...
+        handlers[name](args).then(
+          result => {
+            ipc.server.emit(
+              socket,
+              'message',
+              JSON.stringify({ type: 'reply', id, result })
+            )
+          },
+          error => {
+            // Up to you how to handle errors, if you want to forward
```
okdistribute: TODO: handle errors

src/main/mapeo.js
```diff
@@ -0,0 +1,248 @@
...
+const os = require('os')
+const Settings = require('@mapeo/settings')
+const sublevel = require('subleveldown')
+
+const TileImporter = require('./tile-importer')
+const createServer = require('./server')
+const installStatsIndex = require('./osm-stats')
+
+// This is an RPC wrapper for all the things related to mapeo/core
+// much of this code probably could instead be in mapeo/core
```
okdistribute: to note again: much of this code probably could instead be in mapeo/core

src/main/server.js
```diff
@@ -43,152 +33,8 @@ module.exports = function (osm, media, sendIpc, opts) {
     }
   })
 
-  server.mapeo = mapeoCore
-
-  const origListen = server.listen
-  const origClose = server.close
```
okdistribute: This code was copy and pasted to mapeo-rpc

Issue #261, opened 2019-12-16
Remove or fix "Import map tiles" function from menu


aliya-ryan:
Not really a bug but this feature doesn't currently work so best remove it until it's fixed so users don't get confused (or fix it...).

Team knows about this but I wasn't sure if it was written down so logging here.

okdistribute:
This was fixed in #262 if the user follows [the guide here](https://digital-democracy.gitbook.io/mapeo/custom-imagery), but I think there needs to be more extensive work done on how to make this easier for users.

PR #262, opened 2019-12-17
fix: Remove asar from the import menu

Nmargolis:
### Context
On a mac, when you try to import custom tiles, the option for using tar files was not visible in the UI. Since asar compatibility will be [deprecated soon](https://github.com/digidem/mapeo-server/issues/53), this PR removes it from the default menu options.
The tar format will now be the default format for uploading tiles, and this, combined with the [tilserver port change in id-mapeo](https://github.com/digidem/iD-mapeo/pull/9) fixes this issue: https://github.com/digidem/mapeo-desktop/issues/261

cc @gmaclennan @okdistribute


okdistribute:
failures are unrelated to this PR. thanks @Nmargolis !!

PR #267, opened 2020-01-23
View & Edit observations in Territory View

okdistribute:
This PR is for easily testing the new 'observation editor' feature in the Territory view.

## Approach
iD editor has a lot of custom logic for rendering and managing nodes that isn't that easy to untangle or generalize for observations and nodes. I took a chance on it anyway and [spent a day trying to untangle it](https://github.com/digidem/iD-mapeo/pull/10), but kept finding undocumented and unknown places where things were breaking, and observations were not being rendered without changing their type to 'node'.I think we'll have considerably less code by editing the osm-p2p-server to do the transformation for us.

When iD editor requests the osm data through `/api/0.6/map?bbox=-90,-90,90,90` command, we have a module `obj2osm` which handles turning the objects from the kappa-osm database into xml entities conform with the OSM specification. In this approach, I changed observations to type `node` in the XML.

When a user edits the observation, it creates a changeset and sends it to `osm-p2p-server`. This module now checks if the node's version is pointing to an observation. If so, it patches that observation with the new tags and keeps attachments (and other observation-specific metadata) intact.

## Known issues
- [x] Editing in Territory view, then switching back to Observations view doesn't show edits because the Observations view isn't getting re-rendered. Refreshing then shows the edits. 

## Relevant PRs
- https://github.com/digidem/osm-p2p-server/pull/47


gmaclennan:
@karissa the XML records should always have a `version` prop, which we could use to look up whether it was an observation, then we could ensure that other properties of the observation (`attachments`, which aren't stored under `tags`, and `metadata`) which are ignored / dropped by iD are merged into the edit. That way we could avoid the `__mapeo_type` tag.

okdistribute:
is this ready for merge? waiting on @gmaclennan approval

gmaclennan:
@karissa I rebased this on current `master`, so you will need to reset your local branch.

This is looking really good, combined with the other work on moving Mapeo server into its own process. I've been testing out different data editing scenarios with sample data from our partners and its working well.

I think this is good to merge, but there are some other changes I want to make before we release with this feature:

- [ ] Patch translations of presets into iD translation files - currently these are added when iD is built, but we need to load them from settings files and patch them in.
- [ ] Patch in `{ tags: { categoryId: `{presetId}` } }` to all presets that are passed to iD Editor - this will allow the observation points to match the correct preset.
- [ ] Patch in the default presets needed by iD Editor (`point`, `line`, `area`, `relation`) and the default `name` field to the presets before they are passed to iD, because missing these causes iD to crash, which is confusing for users, and adding them to the presets in mapeo-settings-builder causes these superfluous categories to show up on Mobile.

We can maybe merge this as is and add those in a separate PR. I reviewed the osm-p2p-server PR and it looks good to go too.

okdistribute:
@gmaclennan If you think these are blocking the release, then let's not merge it into master. Ideally the master branch would be only contain features we're ready to release. 

Do you want me to tackle these issues?

gmaclennan:
@okdistribute This is all done now and ready to merge

PR #277, opened 2020-02-07
fix: Fix Windows CI/CD for automated builds

okdistribute:
No idea why rimraf v3 broke this, but pinning to v2 seems to work


noffle:
Nice catch!

PR #284, opened 2020-02-12
fix: only one window at a time during startup.

### LINKED ISSUE #283: When Mapeo loads, there's a big white screen
**Describe the bug**
When Mapeo loads, there's a big white screen and it looks like there is something wrong with it.

Yes, this happens every time you open Mapeo.

**Expected behavior**
The splash screen should replace the empty white screen and the white screen should be invisible upon load.

**Desktop:**
 - OS: All OSs

okdistribute:
fixes #283

### Contributor checklist:

### Description

Tested this on my linux laptop.

This bug was introduced with the new window state manager.

### DIFF
--- a/index.js
+++ b/index.js
@@ -282,6 +282,8 @@ function notifyReady (done) {
       var IS_TEST = process.env.NODE_ENV === 'test'
       if (IS_TEST) win.setSize(1000, 800, false)
       if (argv.debug) win.webContents.openDevTools()
+
+      win.maximize()
       splash.destroy()
       win.show()
       done()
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -282,6 +282,9 @@ function notifyReady (done) {
       var IS_TEST = process.env.NODE_ENV === 'test'
       if (IS_TEST) win.setSize(1000, 800, false)
       if (argv.debug) win.webContents.openDevTools()
+
+      win.maximize()
+      win.setFullScreen(true)
```
marcoautiero: I would avoid to call win.setFullScreen(true)
it sets the app...fullscreen (here on ubuntu) and I don't think is the desired behaviour

okdistribute: You're right good catch! thanks!

PR #303, opened 2020-03-11
New Report Generator

okdistribute:
This branch renders reports as PDFs and removes the dependency on `react-mapfilter`

This is a large PR, I know, but all of the relevant NEW code is quite small, with  logic contained within:
* `src/renderer/components/MapFilter/ReportView/ReportView.js`.

 The rest of the changes are just copy-pasted from `react-mapfilter#pdf-reports` branch.

# TODO
- [x] Remove unnecessary console.logs and replace with logger.log
- [x] Get ReactPDF to render in electron-renderer
- [x] Prev/next buttons
- [x] Styling
- [x] Print button
- [x] Codesplit the report view
- [x] Render a mini map in each page 
- [x] Render a point on the minimap for each observation

# Bugs
- [x] Images are not being lazy-loaded which causes reports to take a long time to load the first time they are generated.
- [x] If you click forward too quickly, the Report view gets 'stuck' and won't advance anymore, you then have to click 'previous' then 'next' again to continue viewing

# Cleanup 
- [x] CSS issue with prev/next buttons
- [x] Factor out hidden-mapbox into it's own module
- [x] Loading indicator after filter is applied and before pdf has rendered 

# v2
- [ ] Add intro page with a map of all observations and a summary of the observations in report (see prototype given by @jencastrodoesstuff)
- [ ] Preview more than 5 pages at a time


gmaclennan:
Hi @okdistribute I pushed my work on this but there is still work to do. I fixed the issue with React PDF not rendering in electron by switching the Webpack config to bundle all of the modules in `node_modules`. This will impact the overall app bundle size because the JS bundle `static/bundle.js` is now 11mb (production build) vs. <1mb, but we are still bundling all the `node_modules` which are no longer needed. Once we fix that duplication, the app size should be much smaller than before. Working from a bundle should also speed up initial load of the app. On my machine it goes from about 5600ms to 1400ms (production build) and it might be even better on spinning-disk machines.

After fixing the bundling the next issue was errors with React Context. It seems like React PDF was unable to deal with any component that read from context (it throws an error). The main uses for context were the "react-pdf=true" context and also the <Fomatted> components from react-intl.

After fixing those I ran into a bug that was unable to render some of the `FeaturePage` code that was returning an array as a sibling to other View components. I wrapped the array in a `<View>` and it fixed it.

After that bug I hit a bug of React PDF not being able to render components which return `null`. I had to stop before fixing that.

I understand it that you were able to render to a PDF with the code when you were running it all on the server? I wonder if all these bugs are in `PDFViewer`, which is rendering everything to the DOM rather than a PDF for a preview. I don't think we need to use this - electron should be able to show a PDF in an iFrame, so we should be able to render to the PDF blob and show it directly, but I haven't tried that yet. For performance it might be worth only rendering one page of the PDF at a time.

I can continue to look at this next week - I feel like I have got my head into the code now and I can make some progress.

gmaclennan:
Note: to use the Chrome default PDF viewer in Electron we will have to upgrade to Electron v9 beta (set to be released as stable on May 19th). See https://github.com/electron/electron/issues/12337

gmaclennan:
Made some good progress today. Confirmed that many of the bugs are due to bugs in [PDFViewer](https://react-pdf.org/components#pdfviewer) from `@react-pdf/renderer`. Found an alternative, to render the PDF using [PDF.js](https://github.com/mozilla/pdf.js/) via [`react-pdf`](https://github.com/wojtekmaj/react-pdf). `PDF.js` is the default PDF viewer in Firefox so should be stable and work well. It gives us much more control over page and print controls than using the Chrome PDF plugin (which would show it's own print, save and zoom controls).

I got storybook working again (some modules specific to electron needed to be mocked/stubbed) and was able to get the report rendering within storybook.

The main barrier to getting it working in the electron app now appears to be getting the `PDF.js` worker loading correctly. This seems to be related to both the webpack build config we use conflicting somehow with the way `react-pdf` (not `@react-pdf`) loads the worker, and it is also likely due to security restrictions on `file://` URLs, so we might need to use [`electron-serve`](https://github.com/sindresorhus/electron-serve).

Also probably performance work is needed for the preview when the user has selected many (dozens, hundreds) of observations. Because with `react-pdf` we can control the page turner, we can actually render a PDF of only the prev / current / next observations and then re-render as the user turns pages. Will check how perf is without it, but images and likely the map will slow things down if we try to render the entire report on every filter / config change.

gmaclennan:
Some more progress today. I tracked down the PDF viewer bug to react-pdf using an older version of PDF.js prior to [a fix for electron](https://github.com/mozilla/pdf.js/commit/1472c10bab3aa71a6efa9c70c6971cfb03dba006#diff-b97ac7a8f6482bf32351140e3e0a2618). Switching to `react-pdf@5-beta.3` has fixed it and I can view PDFs in Mapeo Desktop :) Currently limiting it to 50 pages and switching filters is reasonably fast. Can't push yet because I changed a bunch of other things when I was trying to fix it, and I need to undo all of those now.

image

okdistribute:
nice work @gmaclennan !

jencastrodoesstuff:
Amazing!!!
Round of applause!

*Jen Castro*
*(Remote from Montreal)*
*Program Coordinator - **Digital Democracy*
Tel Perú:  +51(1)7007968 , Canada: +16479455252 , WhatsApp: +17789984058,
Skype:  live:jen_5964

**Pido disculpas por cualquier error gramática u ortográfica. Estoy en el
proceso de mejorar mi nivel de castellano y aprecio su paciencia.

okdistribute:
The prod build fails `npm run build` with

```
bugsnag.js:1724 [bugsnag] Uncaught exception, the process will now terminate…
null
(anonymous) @ build.js:94
(anonymous) @ build.js:94
o._notify @ build.js:94
Qe @ build.js:94
emit @ events.js:228
window.onerror @ electron/js2c/renderer_init.js:114
4854009274ff2f49919f.worker.js:1 Uncaught ReferenceError: process is not defined
    at 4854009274ff2f49919f.worker.js:1
```

okdistribute:
I merged in latest changes from master and getting a new React error related to fonts.

okdistribute:
I don't understand why github is confused about this

patches/@react-pdf+layout+2.0.0-beta.14.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@react-pdf/layout/lib/text/getAttributedString.js b/node_modules/@react-pdf/layout/lib/text/getAttributedString.js
```
okdistribute, line 1: @gmaclennan have you sent a PR for this yet to the module?

package.json
```diff
@@ -109,6 +109,7 @@
     "electron-timber": "^0.5.1",
     "electron-updater": "^4.2.2",
     "flat": "^5.0.0",
+    "flow-geojson": "^2.0.7",
```
okdistribute, line 112: is this still needed? (and `@babel/preset-flow`?) @gmaclennan

gmaclennan: These are both needed for the type checking (we use types from flow-geojson and babel preset allows linting and compile to work on code with flow.)

src/background/cache.js
```diff
@@ -0,0 +1,240 @@
+const crypto = require('crypto')
```
okdistribute, line 1: TODO: Split this file out into it's own module (from express-asset-file-cache-middleware)

src/background/maps.js
```diff
@@ -0,0 +1,47 @@
+var Router = require('routes')
+var mapStream = require('mapbox-map-image-stream')
```
okdistribute, line 2: This uses mapbox-gl under the hood which seems to use the native version

gmaclennan:
Great work on this @okdistribute. As discussed on our check-in today, I think it might make sense to just use the code in https://github.com/gmaclennan/mapbox-gl-local-static-maps/blob/master/index.js for generating map images. `mapbox-map-image-stream` has a bit of overhead due to the way it does a stream of the map image and mosaics tiles from multiple map instances. It does this in order to handle extra large hi-res maps (there is an upper limit on the map canvas size, and memory use would shoot up otherwise) but it makes the map generation slower. Since for reports we are not going to need map images larger than about 2000px for an A4/Letter report, I think it will be fine to just use `map.getCanvas().toBlob()`. The [`getMapImage()`](https://github.com/gmaclennan/mapbox-gl-local-static-maps/blob/master/index.js#L52) function should have all you need.

For caching, another option is to lean on browser caching with a `Cache-Control: max-age` header. The browser (e.g. the Chrome renderer process where the report is being generated) will take care of caching the map images, so we do not need the file cache module.

The style and access-token are best managed by the renderer process (they are already defined there) and passed as query parameters (I imagine what you have currently is just for testing). For the API HTTP end-point, for the mapbox-gl-local-static-maps lib I used the same API as the [Mapbox Static Maps API](https://docs.mapbox.com/api/maps/#static-images). It wouldn't be my first choice of API design, but it seems to make sense to use something compatible with a defined API. To parse the URL I [statically generate a RegExp based on a path expression](https://github.com/gmaclennan/mapbox-gl-local-static-maps/blob/master/lib/generate-regexp.js) and then use [`parseMapOptions`](https://github.com/gmaclennan/mapbox-gl-local-static-maps/blob/master/sw.js#L153-L191) to parse the URL. I'm not wedded to this API, but sharing here in case this code is helpful for you.

I do think it makes most sense to define the API with a center and zoom, rather than a bounding box, because it's easier to generate a bounding box from center,zoom,dimensions than vice-versa.

For showing observations on the map there are a few options:

1. For the maps for each individual observation, the cheating / easy way is to just overlay a marker in the middle of the image (since the map center will be the observation location, and we only need to show a single observation on that map).
2. For the overview map that shows all observations in the report (if we choose to have one, this could be "reports v2"), I see two options:
    1. Encode the overlay / markers into the URL, like the [Mapbox Static Maps API](https://docs.mapbox.com/api/maps/#overlay-options) does. This means that our static maps server does not need to know anything about observations or the current filter, everything it needs to know is encoded in the URL. I feel good about being able to write this in half a day if you don't have the bandwidth for it.
    2. Pass the current filter encoded in the URL, and then the map server requests the observations and filters them on the server side, then adds them as a layer to the map. The request for a list of all observations from osm-db can be slow though, so I think this might be a perf issue, and more complicated than (i) to implement.

okdistribute:
We probably need a bounding box for the front page, but zoom/center for each observation seems fine

PR #315, opened 2020-03-20
Show Wi-Fi SSID

thibautRe:
### Contributor checklist:

- [ ] My contribution is **not** related to translations. _Please submit translation changes via our [Mapeo Crowdin project](https://crowdin.com/project/mapeo-desktop)_
- [x] My commits are in nice logical chunks with [good commit messages](http://chris.beams.io/posts/git-commit/)
- [x] I have walked through the [QA Manual Testing
  Script](https://github.com/digidem/mapeo-desktop/blob/master/docs/testing.md) and updated it if necessary.
- [x] If my changes depend upon an update to a dependency, I have updated the package-lock.json file using `npm install --package-lock`
- [ ] My changes have been tested with the [mobile app](https://github.com/digidem/mapeo-mobile/releases). 
- [ ] My changes are ready to be shipped to users on Windows, Mac, and Linux

Fixes #301 

### Description

This PR adds an information on the SSID in the **Synchronize** view. It is using the `node-wifi` NPM package.

| Connected | Disconnected |
| --- | --- |
| Capture8 | Capture9 | 

### NOTE

Opening this as a draft PR for a few reasons:
- I haven't ran the automated tests yet
- I have only tested the feature on Windows - I will try it on a Mac tomorrow but if anybody can try a Linux build that would be great
- I have touched the translations (both added the "Disconnected" string, as well as removed the "via Wi-Fi" one) but haven't really digged into how the translation system works. I've also seen on Slack that there was some work planned on changing the translation process at some point soon - so anyway if I can get some guidance on how to sync translations strings in the project that would be great :)

Issue #67, opened 2016-12-01
Tool: Smooth line drawing


aliya-ryan:
We need a tool that will create smoother more rounded lines for rivers, lakes etc.

PR #316, opened 2020-03-22
feat: Show WiFi strength indication

### LINKED ISSUE #302: Show WiFi signal strength in sync screen
**Describe the idea**
Show a warning to users if WiFi signal strength is weak, since this will affect sync performance.
For more context see https://github.com/digidem/mapeo-mobile/issues/331

[`node-wifi`](https://github.com/friedrith/node-wifi) should provide this info.

Related issue: #301

thibautRe:
### Contributor checklist:

- [ ] My changes are ready to be shipped to users on Windows, Mac, and Linux

### Description

Fixes #302 

This PR is build on top of #315 - and as such should probably be reviewed only after. **The actual diff between the previous PR and this one [can be accessed here](https://github.com/thibautRe/mapeo-desktop/compare/feat/wifi...thibautRe:feat/wifi-strength)**

I've added a color-coding indication on the WiFi symbol where it is **red** below 30%, **orange** below 60% and black above that. I actually don't know what values are common to encounter in the field (in my whole appartement I couldn't go below 80%) so I've just divided them equally. I also added a tooltip for the user to easily access the current precise signal quality value.

I've added a `Math.min(100, currentConnection.quality)` because in my case I had some signal quality of 120% in my living room.
 
mapeo-wifi-quality


src/renderer/components/SyncView/SyncAppBar.js
```diff
@@ -1,15 +1,29 @@
...
 import Toolbar from '@material-ui/core/Toolbar'
 import Typography from '@material-ui/core/Typography'
 import Button from '@material-ui/core/Button'
 import { makeStyles } from '@material-ui/core/styles'
+import { Wifi, WifiOff } from '@material-ui/icons'
 import { defineMessages, useIntl } from 'react-intl'
+import wifi from 'node-wifi'
+import { Tooltip } from '@material-ui/core'
+
+wifi.init({
```
okdistribute: Is this OK to be doing this in the renderer process? I wonder if it will visibly block it from refreshing?

We could let this go in to master for now and I can rebase it onto my branch for putting a lot of the code into a background IPC process (#259) which would ensure it would not block the renderer. This would require line 40 to be using the ipc to fetch the current connections, but otherwise the code shouldn't have to change.

gmaclennan: I can't see why any of these methods would be processor intensive, so I don't think they would affect the renderer.

thibautRe: The init function seems to just be doing some options initialisation and a bit of currying : the functions like `linuxConnect` in the `switch/case` have a prototype like `(config) => () => {}` so they don't execute anything.

thibautRe: https://github.com/friedrith/node-wifi/blob/master/src/wifi.js#L20

okdistribute:
Hi @thibautRe, this is great and we want to merge. There are two changes that our team would like to see, if you have time. If you don't have time, feel free to let us know and I will take a look at finishing it up!

thibautRe:
Thank you for the feedback @okdistribute ! Just so I'm sure that I understood what you have in mind : you would like to inline the current popover content next to the WiFi symbol or name (and/or reformulate it), and instead put some action tips in the popover. Is that correct?

I will try to get something done by tomorrow :)

thibautRe:
I pushed some changes, here's how it looks now:

Screenshot 2020-03-31 at 22 16 09

And here are the 3 main signal levels:

| Poor signal | Weak signal | Good signal |
| --- | --- | --- |
| Screenshot 2020-03-31 at 22 14 35 | Screenshot 2020-03-31 at 22 16 00 | Screenshot 2020-03-31 at 22 18 34 |

Let me know if you feel like this is closer to what you have in mind or if you would like another improvement :)

src/renderer/components/SyncView/SyncAppBar.js
```diff
@@ -7,15 +7,37 @@ import { makeStyles } from '@material-ui/core/styles'
...
   iface: null
 })
 
+const getQualityStyle = (connection) => {
+  if (connection.quality < 30) return { color: 'white', backgroundColor: 'red' }
+  if (connection.quality < 60) return { color: 'white', backgroundColor: 'orange' }
+  return { backgroundColor: '#E0E0E0' }
+}
+
+/** @returns {keyof m} */
```
thibautRe: This little fella is useful for autocompletion:
Screenshot 2020-03-31 at 21 47 46

okdistribute: This does not bother me, thanks for that! What plugin is this using in VS Code?

thibautRe: None, it's the default Intellisense (or whatever they call that) thingie included in VSCode. It's using JSDoc comments that [Typescript works with and understand](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#supported-jsdoc)

src/renderer/components/SyncView/SyncAppBar.js
```diff
@@ -29,33 +51,47 @@ const SyncAppBar = ({ onClickSelectSyncfile, onClickNewSyncfile }) => {
   // Check connection every 2 seconds
   useEffect(() => {
     const intervalCheck = setInterval(() => {
-      wifi
-        .getCurrentConnections()
-        .then((conn) => setCurrentConnection(conn && conn[0]))
+      wifi.getCurrentConnections().then((conn) => setCurrentConnection(conn && conn[0]))
```
gmaclennan, line 54: Can you add a `catch` for errors here? Probably ok to fail silently, or maybe add a message like "connection unknown".

src/renderer/components/SyncView/SyncAppBar.js
gmaclennan: That's a really great tip @thibautRe. I knew TS could support JS-Doc but I never got round to reading up how - good to know there is nothing to do!

src/renderer/components/SyncView/SyncAppBar.js
```diff
@@ -7,15 +7,37 @@ import { makeStyles } from '@material-ui/core/styles'
...
 })
 
+const getQualityStyle = (connection) => {
+  if (connection.quality < 30) return { color: 'white', backgroundColor: 'red' }
+  if (connection.quality < 60) return { color: 'white', backgroundColor: 'orange' }
+  return { backgroundColor: '#E0E0E0' }
+}
+
+/** @returns {keyof m} */
+const getWifiConnectionMessage = (connection) => {
```
gmaclennan, line 23: Just a point of style - fine to ignore - but this might make more sense to return the message object e.g. `m.qualityTooltipPoor`, then line #67 can be `t(getWifiConnectionMessage(currentConnection))`.

thibautRe:
Thank you for the feedback! Fixed all the comments, here's how it would look in the errored state:

Screenshot 2020-04-01 at 18 41 42

src/renderer/components/SyncView/SyncAppBar.js
thibautRe: Makes sense! Then I can get rid of the JSDoc comment as well (it lived a short but intense life)

okdistribute:
awesome looks great! Thanks @thibautRe

PR #329, opened 2020-04-17
feat: Use production build of react in packaged app

gmaclennan:
Currently we are using the development build of react in the packaged
app. The development build is slower, see:
https://reactjs.org/docs/optimizing-performance.html

The fix is a little complicated: Normally you would build a bundle (including React) with `process.env.NODE_ENV = 'production'`. However, we use webpack with `target='electron-renderer'` and this loads all modules as externals e.g. they are required at runtime rather than included in a single bundle. The `webpack.environmentPlugin()` does not set `NODE_ENV` for runtime, it only replaces the variable in moduels included in the bundle (react isn't included, because it is 'external').

The hacky workaround is to patch in `process.env.NODE_ENV = 'production'` to the output bundle from webpack. A better long-term solution might be to bundle all code for electron, which might also result in a smaller build size and faster loading (lots of `require` calls has an overhead, especially on slow machines with spinning disk hard drives). This would require filtering out dependencies from the packaged app which are required in the bundle.

TL;DR this change should make the app more responsive (snappier™) and should work well enough for now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "mapeo-desktop",
-  "version": "5.0.7",
+  "version": "5.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


okdistribute:
@gmaclennan Thanks for this PR! We want to keep the package-lock.json file though, unless I'm missing something. Was that a mistake? Could you add it back? Thanks!

gmaclennan:
@okdistribute it's still there, I see it in the "Files changed" tab. It does have quite a few changes. I'm not sure how to install an additional module without changing it.

okdistribute:
Sigh, package-lock.. why..

PR #330, opened 2020-04-17
fix: Improve conversion of Mapeo presets for iD Editor

gmaclennan:
The schema for Mapeo presets and fields differs slightly from iD Editor. This improves the conversion of presets used in Mapeo so that they better appear when viewing monitoring data in the territory view of Mapeo Desktop.

- Fixes fields of type `text` where `appearance=singleline` (was causing crash)
- Uses the iD Editor `semiCombo` type to display select-multiple fields (NB: The resulting field is a semi-colon-separated string, whereas Mapeo expects this to be an array, so this will not work for editing observations in the territory view).
- Uses the correct labels defined in the preset for select fields.

### DIFF
--- a/src/renderer/components/MapEditor/index.js
+++ b/src/renderer/components/MapEditor/index.js
@@ -377,14 +377,25 @@ function convertPresets (presetsObj) {
     let type
     let snakeCase
     switch (field.type) {
+      // The iD equivalent of the mapeo select_one is a `combo` field
       case 'select_one':
         type = 'combo'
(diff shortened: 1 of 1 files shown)


okdistribute:
I'd love to try to make this into v5.1.1 - Any timeline on when you think this will be ready? I can also take a look at finishing it up soon if you have limited time.

gmaclennan:
I think this is good to merge - wanted to check it with fresh eyes in the morning. The issue with multi-selects exists prior to this PR, and we can fix that separately. I've opened an issue #334

PR #333, opened 2020-04-17
chore: ensure website is only released from tags on master branch

okdistribute:


.github/workflows/nodejs.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           WIN_CSC_LINK: ${{ secrets.WIN_CSC_LINK }}
           WIN_CSC_KEY_PASSWORD: ${{ secrets.WIN_CSC_KEY_PASSWORD }}
       - name: Update website release
-        if: contains(github.ref, '/tags/')
+        if: contains(github.ref, '/master/') && contains(github.ref, '/tags/')
```
gmaclennan, line 56: This won't work unfortunately. When this is triggered by a tag, the ref will be the tag, not the branch (see https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context). I'm not entirely sure how to solve this.

okdistribute: Hmm, we could then check if we're on the master branch inside the script, maybe? I'll look at this next week.

.github/workflows/nodejs.yml
```diff
@@ -57,7 +57,10 @@ jobs:
           WIN_CSC_LINK: ${{ secrets.WIN_CSC_LINK }}
           WIN_CSC_KEY_PASSWORD: ${{ secrets.WIN_CSC_KEY_PASSWORD }}
       - name: Update website release
-        if: contains(github.ref, '/tags/')
+        if: contains(github.ref, '/tags/') && github.event_name === 'push'
+        on:
```
okdistribute, line 61: not sure if 'on' works at this level

okdistribute:
Had a shower thought that this could be in it's own workflow, which would allow `on.branch.master` and `contains(github.ref, /tag/)`

okdistribute:
We will instead release to the website manually

PR #343, opened 2020-05-07
Chore: use package lock

gmaclennan:
I noticed that we had an `.npmrc` file that was turning package-lock off, and we were using `npm install` on Github CI which does not strictly follow package-lock.

This PR turns package-lock on, and uses `npm ci` to install things on CI. We should test that the builds from this PR work before merging.


okdistribute:
awesome catch!

Issue #359, opened 2020-06-10
Sync file not finishing


okdistribute:
**Describe the bug**
From Jen
"The wheel is spinning but the numbers haven’t changed. I’m lost on how to offer you valuable information to figure out what is or isn’t going on."

**Can you do the same thing twice?**
1. In the shared drive, download the ECA recuperando 20 May `.mapeodata` file
2. Syncronize from a syncfile
3. See it spinning

**Expected behavior**
It should finish syncing

**Screenshots**
<img width="351" alt="Screen Shot 2020-06-10 at 1 51 36 PM" src="https://user-images.githubusercontent.com/633012/84322838-8a91ab80-ab2a-11ea-8539-26df7c387259.png">

**Desktop:**
Mac
v5.2.1

okdistribute:
Was able to get it into this state today 
![Screenshot from 2020-06-10 17-15-54](https://user-images.githubusercontent.com/633012/84331187-68a22400-ab3e-11ea-989a-88c8c5154c08.png)

okdistribute:
This should be fixed in 5.3.0

PR #360, opened 2020-06-10
Cleanup worker child processes in production

okdistribute:
This will solve one of the issues reported in #359 -- when the user is forced to 'force quit' Mapeo due to the sync screen getting stuck and never finishing, the child process was being a zombie on the computer, and @jencastrodoesstuff reported having to reinstall Mapeo or restart the computer in order to get it to work as intended again.

It also will make the sync screen go a bit smoother, there was lots of logging happening that was slowing it down considerably. We need to create a debug log for the sync screen which can be enabled separately only when a developer opts-in, or when there are critical errors.


noffle:
Hey! I poked at this a bit, and here's what I learned:

- `capture-exit` works by hooking into Node's `process.on('beforeExit')` event. This only fires if the shutdown is graceful. Signals like SIGTERM,SIGINT,SIGSEGV,etc will prevent this event from firing.
- `signal-exit` looks like it will capture unix signals directly. I tested this locally, and it will capture things like SIGINT and SIGTERM ok, but not segmentation faults or floating point errors. According to the module docs, it *could* catch them, but the process state is too undefined to run and JS. SIGKILL seems to be uncatchable.
- nitpick: [zombie processes](https://en.wikipedia.org/wiki/Zombie_process) are processes that have completed but not yet cleaned up. [orphan processes](https://en.wikipedia.org/wiki/Orphan_process) are ones where the parent has died but the child is still alive.
- I wonder if we could additionally have Mapeo check for an old orphan process that didn't get cleaned up, and SIGTERM it?

gmaclennan:
PS. I agree with Kira's nitpick, but I prefer a PR title that includes "Kill zombies".

okdistribute:
- Create a pid file
- kill pid & delete on shutdown
- if the pid file exists on startup, kill pid

index.js
```diff
@@ -30,12 +35,27 @@ debug({ showDevTools: false })
...
+// Before we do anything, let's make sure we're ready to gracefully shut down
+var serverProcess = null
+const onExit = require('capture-exit')
+const signalExit = require('signal-exit')
+onExit.captureExit()
+onExit.onExit(() => {
+  killZombies()
+})
+
+signalExit((code, signal) => {
```
okdistribute, line 49: @noffle I don't know if we need both of these at this point, capture-exit works for graceful exits during async calls and signal-exit is for external sig* sends...

src/pid.js
```diff
@@ -0,0 +1,38 @@
...
+  constructor (userDataPath) {
+    this.loc = path.join(userDataPath, 'pid')
+  }
+
+  remove () {
+    return fs.unlinkSync(this.loc)
+  }
+
+  exists () {
+    return fs.existsSync(this.loc)
```
okdistribute, line 16: I figure it's OK to block because if there is a pid we don't want to do anything until we delete it and restart the process

index.js
```diff
@@ -330,20 +347,29 @@ function beforeQuit (e) {
...
-    if (serverProcess) serverProcess.kill()
     try { closingWin.close() } catch (e) {}
     closingWin = null
-    serverProcess = null
+    killZombies()
     app.exit()
   })
 }
 
+function killZombies () {
```
okdistribute, line 360: I know it's technically an orphan but I feel bad about killing them

index.js
```diff
@@ -44,6 +64,7 @@ if (!gotTheLock) {
   // Didn't get a lock, because another instance is open, so we quit
   process.exit(0)
 } else {
+  if (pid.exists()) pid.kill()
```
okdistribute, line 67: todo: add callback before forking process

index.js
noffle: Computer scientist terminology, huh? Maybe we `resolve` the orphan? :P

src/logger.js
```diff
@@ -1,9 +1,13 @@
 const winston = require('winston')
 const path = require('path')
 const DailyRotateFile = require('winston-daily-rotate-file')
-const store = require('./store')
 const util = require('util')
 
+const store = require('./store')
+const appVersion = require('../package.json').version
+
+const BUGSNAG_API_KEY = 'fcd92279c11ac971b4bd29b646ec4125'
```
noffle, line 9: Will this let anybody make API requests to bugsnag? Could randos use this to read sensitive reports?

index.js
noffle: Oh, if signal-exit doesn't fire on graceful exits, then we'd need both to capture graceful & nongraceful exits, right?

src/pid.js
noffle: +1

src/logger.js
okdistribute: This will let anyone make API requests to bugsnag. I'll hide it in the source code using an environment variable. It won't let randos read sensitive reports.

index.js
```diff
@@ -331,20 +336,16 @@ function beforeQuit (e) {
...
   ipc.send('close', null, () => {
-    logger.info('IPC closed')
-    clearTimeout(closingTimeoutId)
-    if (serverProcess) serverProcess.kill()
-    try { closingWin.close() } catch (e) {}
-    closingWin = null
-    serverProcess = null
-    app.exit()
+    logger.debug('IPC closed')
+    worker.cleanup((err) => {
```
okdistribute, line 342: @noffle now we are cleaning up the worker ;P

index.js
okdistribute: yeah. in any case, they will only ever be called once!

Issue #324, opened 2020-04-04
Integrate bugsnag


okdistribute:
**Describe the idea**
Integrate bugsnag into the application so that we can get a better picture of when fatal crashes happen.

**What's your role? Where and how do you currently use the application?**
Lead on Mapeo Desktop @ Dd

**What is your ideal step by step workflow?**
All fatal application crashes and uncaught errors should be sent to bugsnag and also logged on the device in a way that can be retrieved at a later date. These should include very minimal information -- a timestamp, the module the error occurred in, the stack trace, and perhaps some of the local variables. However, we have to ensure that the variables that are logged will not open up user data to unwanted surveillance. No sensitive user data, just non-identifiable internal application state, should be logged.

**Have you thought about what it might look like?**
It should also open an issue on the GitHub repository once an issue has been reported.

okdistribute:
fixed by #361

okdistribute:
Fixed in 5.3.x

PR #361, opened 2020-06-15
Better logging, and log files written to disk

okdistribute:
Writes files to disk, and improves logging with three levels:

* debug (for developers), printed only to console
* info (good things, infrequent), printed to console and written to disk in %DATE%.log
* errors (bad things, infrequent), printed to console and written to disk in %DATE%.error.log

All processes (background, main, renderer) will write to these files.

This PR also currently reverts back to old behavior for logging during development debugging, where the background process logs in the background window, the frontend JS in the renderer console, and the electron main process in the terminal.

If we want to keep `electron-timber` behavior that will need to be modified. I have not found it to be not essential for my development, as I prefer the Chrome debugger to the terminal, because I can get nice step-through features. But if other developers are mainly using the terminal and want to see all messages together, I can re-enable that. Thoughts? @gmaclennan @noffle

### DIFF
--- a/src/background/mapeo.js
+++ b/src/background/mapeo.js
@@ -20,13 +20,12 @@ class MapeoRPC {
     this.storages = []
     this.config = new Settings(userDataPath)
     this.encryptionKey = this.config.getEncryptionKey()
-    logger.log('got encryptionKey', this.encryptionKey)
+    logger.info('got encryptionKey', this.encryptionKey.substr(0, 4))
     this.tiles = TileImporter(userDataPath)
 
     var feedsDir = path.join(datadir, 'storage')
     var indexDir = path.join(datadir, 'index')
 
-    logger.log('loading datadir', datadir)
     this.indexDb = level(indexDir)
 
     const coreDb = kappa(datadir, {
@@ -57,7 +56,7 @@ class MapeoRPC {
     this.ipcSend = ipcSend
     this.ipcSend('indexes-loading')
     this.osm.ready(() => {
-      logger.log('indexes READY')
+      logger.debug('indexes READY')
       this.ipcSend('indexes-ready')
     })
 
@@ -96,7 +95,6 @@ class MapeoRPC {
 
     this.core.on('error', this._handleError)
     this.core.sync.listen((err) => {
-      logger.log('mapeo-core sync is listening')
       cb(err)
     })
   }
@@ -104,17 +102,19 @@ class MapeoRPC {
   _syncWatch (sync) {
     const startTime = Date.now()
     var onerror = (err) => {
-      logger.error(err)
+      logger.error('sync', err)
+      // We handle errors separately/differently for sync
+      // instead of sending to _handleError
       sync.removeListener('error', onerror)
       sync.removeListener('progress', this._throttledSendPeerUpdate)
       sync.removeListener('end', onend)
     }
 
     var onend = (err) => {
-      if (err) logger.error(err)
+      if (err) logger.error('sync error', err)
       this.ipcSend('sync-complete')
       const syncDurationSecs = ((Date.now() - startTime) / 1000).toFixed(2)
-      logger.log('Sync completed in ' + syncDurationSecs + ' seconds')
+      logger.info('Sync completed in ' + syncDurationSecs + ' seconds')
       sync.removeListener('error', onerror)
       sync.removeListener('progress', this._throttledSendPeerUpdate)
       sync.removeListener('end', onend)
@@ -129,7 +129,7 @@ class MapeoRPC {
   _onNewPeer (peer) {
     this._throttledSendPeerUpdate(peer)
     if (!peer.sync) {
-      return logger.error('Could not monitor peer, missing sync property')
+      return logger.error('sync', new Error('Could not monitor peer, missing sync property'))
     }
     peer.sync.once('sync-start', () => {
       this._syncWatch(peer.sync)
@@ -137,8 +137,7 @@ class MapeoRPC {
   }
 
   syncStart (target = {}) {
-    logger.log('Sync start request:', target)
-    logger.log('sync starting', target)
+    logger.info('Sync start request:', target)
 
     const sync = this.core.sync.replicate(target, {
       projectKey: this.encryptionKey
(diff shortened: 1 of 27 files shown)


noffle:
Cool!

Does this log debug/info/error to separate files? It could be helpful to
have it all inlined together, as in

June 15 2020 14:02:55 [INFO] Mapeo starting
June 15 2020 14:03:27 [ERROR] Missing config file: presets.json
June 15 2020 14:03:29 [DEBUG] Indexes ready!

Having it together could help with figuring out what series of INFO
events led to an ERROR event, for example.

Are timestamps included? Millisecond accuracy could be helpful, if it's
non-trivial to add.

gmaclennan:
Thanks for taking this on Karissa, I've had this on my list of things to do for a while. A few thoughts on this:

1. I would like to integrate bugsnag into a logging module, basically so we do not need to worry about bugsnag in the code, and the logging module decides what to send to bugsnag. Bugsnag has the concept of ["breadcrumbs"](https://docs.bugsnag.com/platforms/javascript/#leaving-breadcrumbs) which I think in JS are automatically captured from `console.log()` but might not be if we log to a file. In React Native I manually set breadcrumbs in random places. The breadcrumbs are super useful for understanding crash reports which reach Bugsnag.

2. I agree with Kira about all logs being written to the same file, this would be much easier to parse.

3. I think writing debug logs to a file is also very useful. There have been countless times when remote troubleshooting Mapeo when it would be very useful to have a way to record a debug log from an error in production, and for the user to be able to send that log to us to review.

4. I think it would be ideal if a logging library can accept "levels" as a config, so that the user can turn on "debug" logs if we need to troubleshoot. We would have this off as default for performance and disk space reasons (I once troubleshooted a laptop that had no disk space, and most of it was due to Skype debug logging being turned on and a multi-gigabyte log). Best if it turned itself off after a while (because users forget to disable it).

5. I think logging performance is an important consideration. As you have seen with the sync screen it can have surprising impacts on performance. A little while back I watched Matteo Collina's video on logging: https://www.youtube.com/watch?v=Dnx2SPdcDSU and [Pino](https://getpino.io/#/) seems like the fastest option out there, although not necessarily the best API. Most of the perf discussions are about servers, but I think in our case and running on low-power devices, sometimes with slow disk writes, the perf issues are also relevant to Mapeo

6. We should (non-urgent) review code for costly log operations. This includes things like logging the result of `sync.on('peer')` which includes a `socket.Response` object and is costly to write to the console.

It would be great if the work done on this could also be used for Mapeo Mobile. I have found Bugsnag integration to be really helpful, but I would love to remove all the bugsnag code from Mapeo and keep it all in one module (e.g. a mapeo-logger module) so it is easy to swap it out if needed.

Re. `electron-timber` I don't feel particularly strongly about it. I have found it handy to have all the logs in one place, it's easy to forget to open the hidden window for the ipc process and open the chrome debugger there. I also do like the nice formatting, I find it easier to parse the logs, but I could do without it.

gmaclennan:
Would also be cool to integrate the `debug` module into this, e.g. so that messages from modules that use `debug` could be captured to the same log file. This might require overriding `console.log` globally though, since I'm not sure if `debug` has an option to override the log stream globally (only per-module).

okdistribute:
1. I'll make sure that alll INFO and ERROR logs are written to %DATE%.log. I suspect all ERRORS should also be written to %DATE%.error.log
2. I can add a way to upgrade to writing 'debug' logs for a period of time
3. Adding bugsnag too
4. I can look at also integrating the 'debug' module, not sure how that will work exactly but probably possible!

gmaclennan:
To be clear I'm not even sure how important the `debug` part is. But if really want to get to the bottom of something if might be useful to be able to turn on "really detailed logs" and call `debug.enable('*')` and capture `debug` messages from all required modules in addition to Mapeo.

okdistribute:
This is ready for review, I was not able to get `debug.enable` for the `debug` module printing, there is probably a way to do this by overriding `process.stderr`, but I didn't have time today to figure it out but it probably would be useful.

I've gotten to a good stopping place as all the other features are working, including writing extra logging with `debug` setting for 1 day, if they've been set through `Help->Debug`.

I will add a screencast to show this behavior next week some time.

Woot! Logging!

okdistribute:
closes  #324

okdistribute:
There is one more bug -- electron store does not load inside the child process, so that needs to be fixed.

noffle:
> I was not able to get debug.enable for the debug module printing, there is probably a way to do this by overriding process.stderr, but I didn't have time today to figure it out but it probably would be useful.

Oh! I messed with this recently to debug output. I added this as a third param to `fork`:

```js
  ], {
    env: {
      'DEBUG': '*,-dns-discovery,-discovery-swarm,-discovery-channel'
    },
    stdio: 'inherit'
  })
```

You could probably replace the custom string I have with, like,

```js
  ], {
    env: {
      'DEBUG': process.env.DEBUG
    },
    stdio: 'inherit'
  })
```

gmaclennan:
This is really fantastic @okdistribute, super happy to have this integrated. A couple of small things to improve Bugsnag integration:

1. Manually set [`releaseStage`](https://docs.bugsnag.com/platforms/javascript/configuration-options/#releasestage) because it will default to `development` (because we use localhost). I got a lot of noise in Bugsnag for mobile from errors that came from local dev. We can set the release stage to `development` when in local dev, then turn off sending data to bugsnag in that release stage with the option [`enabledReleaseStages`](https://docs.bugsnag.com/platforms/javascript/configuration-options/#enabledreleasestages)
2. [Leave breadcrumbs](https://docs.bugsnag.com/platforms/javascript/#leaving-breadcrumbs) for log events. Normally Bugsnag would hook into `console.log` but I'm not sure if logging goes through console.log now? We need to check if `bugsnag.leaveBreadcrumb` is a function, because it is not supported in Node, only the browser.
3. This one can be added later (after this PR merges): Add the [react plugin](https://docs.bugsnag.com/platforms/javascript/react/) to bugsnag, which will offer more detailed error reporting for react errors.
4. Which Bugsnag project is this linked to? I only currently see the Mapeo Mobile project under the Digital Democracy Bugsnag account.

okdistribute:
done

okdistribute:
bugsnag doesn't like electron very much.. having trouble getting it to show up in the dashboard

src/background/mapeo.js
```diff
@@ -20,13 +20,12 @@ class MapeoRPC {
     this.storages = []
     this.config = new Settings(userDataPath)
     this.encryptionKey = this.config.getEncryptionKey()
-    logger.log('got encryptionKey', this.encryptionKey)
+    logger.debug('got encryptionKey', this.encryptionKey)
```
noffle, line 23: Can we not log this? Or just log the first 4 hex chars? This is going to be delicate enough that we want to minimize where we expose it.

src/logger.js
```diff
@@ -1,8 +1,165 @@
...
+      // if debugging is on and a new log file is created, turn it off
+      this.debugging(false)
+    })
+    mainLog.on('rotate', () => {
+      // if debugging is on during a rotate, turn it off
+      this.debugging(false)
+    })
+    this.winston.add(mainLog)
+
+    const errorTransport = new (DailyRotateFile)({
```
noffle, line 64: Does this mean that errors are still going to a separate file? (I'm not familiar with how this module works.)

src/background/mapeo.js
okdistribute: debug doesn't get written to disk unless you have `debugging` mode turned on -- can do the first 4

src/logger.js
okdistribute: Both!

okdistribute: this can be turned off if we don't want it by removing the errorTransport lines entirely. Going to merge now

PR #364, opened 2020-06-26
chore(deps): Update @mapeo/core to v8.2.2

okdistribute:
This fixes underlying sync state bugs by integrating updates from Mapeo Core that were finished in May by @noffle.

This will need to be a coordinated [release with Mapeo Mobile v5 ](https://github.com/digidem/mapeo-mobile/pull/358)

Depends upon https://github.com/digidem/mapeo-core/pull/94

Here's a demo of me trying to break it! 

https://www.dropbox.com/s/udeevnpniw1r7pn/fix-that-sync-bug.mp4?dl=0

Open question to @gmaclennan  -- do we want to abstract this sync state client logic into a `@mapeo/client` module to re-use in mobile? Do we want to do this now or later? Ideally the only difference btwn mobile & desktop is the user interface/presentation layer, not the state logic

### DIFF
--- a/src/renderer/components/SyncView/SyncButton.js
+++ b/src/renderer/components/SyncView/SyncButton.js
@@ -16,7 +16,11 @@ const m = defineMessages({
   // Button when sync is complete
   complete: 'Complete',
   // Button to retry sync after error
-  retry: 'Retry'
+  retry: 'Retry',
+  // Disconnected
+  disconnected: 'Disconnected',
+  // Almost done! But progress is inaccurate.
+  finishing: 'Finishing…'
 })
 
 const SyncIcon = props => <FontAwesomeIcon icon={faBolt} {...props} />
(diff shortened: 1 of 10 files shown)


src/renderer/components/SyncView/index.js
```diff
@@ -225,6 +225,12 @@ function getPeersStatus ({
       syncRequests.has(serverPeer.id)
     ) {
       status = peerStatus.PROGRESS
+    } else if (
```
okdistribute, line 228: now that errors are captured correctly, we want to show completion first

src/renderer/components/SyncView/index.js
```diff
@@ -138,8 +138,8 @@ function usePeers (listen) {
           updatedServerPeers.forEach(peer => {
             if (!peer.state) return
             if (
-              peer.state.topic === 'replication-error' ||
-              peer.state.topic === 'replication-complete'
+              (peer.state.topic === 'replication-error' ||
+              peer.state.topic === 'replication-complete') && !peer.connected
```
okdistribute, line 142: If a sync request is sent very quickly after a syncronization is complete, this code was removing that sync request before it even started again. The sync request should only be deleted once it is an error or completion AND the peer has disconnected, which means that the entire lifecycle of the sync has completed.

okdistribute: This likely will not be an issue with mapeo-core v9, however

src/renderer/components/SyncView/SyncButton.js
```diff
@@ -16,7 +16,9 @@ const m = defineMessages({
   // Button when sync is complete
   complete: 'Complete',
   // Button to retry sync after error
-  retry: 'Retry'
+  retry: 'Retry',
+  // Disconnected
+  disconnected: 'Disconnected'
```
gmaclennan, line 21: Is this intended to be shown as a button label?

okdistribute: Sme open questions here regarding how to display disconnected state. Right now it's greyed out, but I think it would be helpful to say 'disconnected' there.. will incorporate that and show a screen shot tomorrow

okdistribute: Screenshot from 2020-07-01 11-42-03

okdistribute: Screenshot from 2020-07-01 11-42-03

Also now '100%...' when not finished yet will show 'Finishing...' instead

PR #370, opened 2020-07-09
fix: Always import default jungle settings

okdistribute:
Reported by @arky that the default settings weren't showing up anymore and he felt this was an important feature as the Observation filter would also stop working in this case.

This is a small fix technically but a big fix for users, as it will always load the jungle presets by default if there is no presets already in `presets/default/presets.json`. 

When we removed the Welcome screen from Mapeo, the default presets stopped being loaded for first-time users. @gmaclennan was this intentional or a bug?

What Mapeo looks like now on first-time setup:

Screenshot from 2020-07-08 22-26-04
Screenshot from 2020-07-08 22-25-48


gmaclennan:
Sorry somehow I managed to loose the review on this (I think I never completed the review once I went into a rabbit hole of fixes).

Mapeo Mobile "falls back" to the default presets, which are distributed with the app via the [npm module](https://www.npmjs.com/package/mapeo-default-settings). The difference with the fallback behaviour is that if we update the default settings distributed with the app, then the user will see the updated settings (if they have never configured custom presets). With the implementation above it would need additional logic to update the default settings.

If possible it would be good to have the same fallback behaviour here. In order to future-proof against problems when we bundle JS code with webpack, the build steps for Mapeo Desktop should copy the default presets folder from `node_modules/mapeo-default-settings` into `static`. I think either `@mapeo/settings` or the code here should implement some kind of fallback behaviour. On MM that is implemented in [`mapeo-server`](https://github.com/digidem/mapeo-server/blob/master/api.js#L145) but has [additional code in the app](https://github.com/digidem/mapeo-mobile/blob/develop/src/backend/server.js#L48-L56) to also fallback to the `projectKey` (which is not yet supported in mapeo-server).

I think the implementation here is _ok_ but we might get stung by it in the future if we need to make some fixes to the default settings.

okdistribute:
@gmaclennan should that code be in mapeo-server or mapeo-core or mapeo-settings modules? I feel really weary about continuing to copy-paste stuff from Mobile, this is how we get out of sync and mobile & desktop don't operate how they should :-\

okdistribute:
@gmaclennan seems like this should be in mapeo/settings `unpackifNew` function -- this function looks at the npm version of the presets that have been imported and updates them if the version is newer. Is there a reason why mobile has custom logic or should this be the same on both?  I feel really weary about continuing to copy-paste stuff from Mobile, this is how we get out of sync and mobile & desktop don't operate how they should :-\

okdistribute:
@gmaclennan it looks like you made another mapeo-default-settings module so we have two? hmm not sure how to proceed here

https://github.com/digidem/mapeo-default-settings/
https://github.com/digidem/mapeo-default-config

okdistribute:
Today I tried using `mapeo-default-settings` but it was too late in my day after debugging sync and I couldn't figure it out because it isn't entirely a straightforward 1:1 with mapeo-styles.

index.js
```diff
@@ -188,8 +189,17 @@ function initDirectories (done) {
   mkdirp.sync(argv.datadir)
   styles.unpackIfNew(userDataPath, function (err) {
     if (err) logger.error('[ERROR] while unpacking styles:', err)
+    var customLocation = path.join(userDataPath, 'presets', 'default')
+    fs.stat(path.join(customLocation, 'metadata.json'), function (err) {
```
okdistribute, line 193: This will unpack if there is not a metadata.json because otherwise it might override it.

okdistribute:
Depends upon https://github.com/digidem/mapeo-styles/pull/3

Ready for review @gmaclennan !

PR #382, opened 2020-07-13
fix: Add 'Get Database Status' menu item

okdistribute:
Depends upon https://github.com/digidem/mapeo-core/pull/98

Screenshot from 2020-07-15 18-37-17
Screenshot from 2020-07-15 18-37-06


src/background/mapeo-ipc.js
```diff
@@ -196,8 +196,8 @@ handlers['get-replicating-peers'] = async () => {
 
 handlers['get-database-status'] = async () => {
   return new Promise((resolve, reject) => {
-    if (!manager.mapeo.core) return reject(new Error('Mapeo must be listening first.'))
-    resolve(manager.mapeo.core.getDatabaseStatus())
+    if (!manager.mapeo) return reject(new Error('Start mapeo before getting active peers!'))
+    manager.mapeo.core.getDatabaseStatus((_, stats) => resolve(stats))
```
gmaclennan, line 200: I know `_` won't be an error in the current implementation, but best to check it and throw and reject here anyway so we don't forget in the future

```suggestion
    manager.mapeo.core.getDatabaseStatus((err, stats) => {
      if (err) return reject(err)
      resolve(stats)
    })
```

gmaclennan:
I'm thinking maybe best not to expose the whole feed key as the id here, partly for readability, also in case it becomes sensitive data in the future. Maybe the first 7 characters? Should be enough for unique ID of any device.

okdistribute:
Depends upon @mapeo/core@8.3.1

okdistribute:
Waiting to merge this until https://github.com/digidem/mapeo-core/pull/100, because it'll require various package-lock bumps in server & desktop

PR #389, opened 2020-07-31
fix: update to electron 9.1.2

### LINKED ISSUE #387: Update to electron >=7
There's a security vulnerability in Electron 6, so we should update to the latest version that also supports node 12.x.

When we upgrade we should also update mapeo/core and related libraries to be automatically tested on the same node version supported by electron.

okdistribute:
fixes #387

### DIFF
--- a/src/renderer/components/MapEditor/index.js
+++ b/src/renderer/components/MapEditor/index.js
@@ -126,7 +126,9 @@ const MapEditor = () => {
       var saved = history.toJSON()
       id.current.flush()
       if (saved) history.fromJSON(saved)
-      ipcRenderer.send('zoom-to-data-get-centroid', 'node', zoomToData)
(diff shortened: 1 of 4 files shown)


src/renderer/components/MapEditor/index.js
```diff
@@ -126,7 +126,9 @@ const MapEditor = () => {
       var saved = history.toJSON()
       id.current.flush()
       if (saved) history.fromJSON(saved)
-      ipcRenderer.send('zoom-to-data-get-centroid', 'node', zoomToData)
+      api.zoomToData('node', (_, loc) => {
```
okdistribute, line 129: This is now throwing an error because you can't send functions as a callback to `ipcRenderer.send` anymore, but we have the same function available in `ipc`

PR #393, opened 2020-08-11
chore: Merge mapFilter@3.2.5 & add bundling & code-splitting

gmaclennan:
Merges [`mapfilter@3.2.5`](https://github.com/digidem/react-mapfilter/tree/v3.2.5) into this repo.

This is the base commit of #392 but moving this into a separate PR so that we can separate out what might break things.

- [x] Storybook working
- [x] react-intl integration + translations from mapfilter
- [x] flow type checking
- [x] prettier code formatting to match mapeo-desktop
- [x] builds work and tested
- [x] mapfilter tests are working
- [x] reduce packaged app size by excluding files in js bundle
- [x] move index.html to support webpack workers in same dir
- [x] bundle split to further improve initial load time
- [x] fix race condition: global.osmServerHost not set at page load
- [x] copy translations from mapfilter into crowdin
- [x] Test single and multi-select fields


gmaclennan:
Bundling all the renderer Javascript improves the initial load time on my machine from ~5.8 seconds. to ~2.3 seconds.

Additional code splitting (splitting out the code for the MapEditor and MapFilter separate from the main bundle) reduces the initial load+parse of JS code to just ~300ms. This means that the user will see the initial Mapeo layout (blue nav bar to the left) approx. 6 seconds earlier (on my fast machine, probably a greater improvement on slower machines) and will see a loading animation while the tab loads, for < 1 second.

Bundling also reduces the Windows packaged app size from 99.6 Mb to 75.9 Mb (24% saving) and Mac builds from 151 Mb to 109.2 Mb (28% saving).

@okdistribute These improvements unfortunately also introduce a race condition: the app loads before the server has started listening, and `global.osmServerHost` is `undefined`. On mobile we use a "status" event listener, which the front-end uses to render loading/error/app. I don't quite understand the way IPC is hooked up, when you are back it would be helpful to look together at the quickest way to solve this race condition.

gmaclennan:
The fix for startup now does not work when you refresh the front-end, because it is waiting for the message from the back-end that never comes (it only fires when the backend starts). We may need a more complex status / heartbeat to get this working reliably.

gmaclennan:
rebased onto master and improved the fix for startup race condition.

gmaclennan:
@okdistribute this is working now on Windows and MacOS. However when I test it in my Linux virtual machines (Ubuntu 18 and Mint 19) when the main window loads it appears transparent (it seems to just capture the desktop image from when it loads):

image

Can you check what you see when running locally from this PR code, and what you see when you install from the artifacts here: https://github.com/digidem/mapeo-desktop/pull/393/checks?check_run_id=999115216

okdistribute:
Do we need to commit the `flow_typed` directory?

gmaclennan:
Yes. I think the way that Typescript does this is better.

okdistribute:
I tested this on Linux and it works, I also fixed the permissions issue in #401 as well as the problem with Select boxes showing the empty string (https://github.com/digidem/mapeo-desktop/pull/393/commits/de922de98d7915588637e0974a34c2325960f0e3)

LGTM to merge!

src/renderer/components/MapFilter/utils/strings.js
```diff
@@ -31,6 +31,7 @@ export function primitiveToString (
   if (typeof value === 'number') return value.toString()
   // TODO: what to show (translated) for "null" field (vs. undefined)
   if (value === null) return 'No Value'
+  if (value === undefined) return undefined
```
gmaclennan, line 34: This will break things, since this function must return a string. However, `value` is type-checked to always be `string | boolean | number | null`.

okdistribute: This is what was causing the error though, Autocomplete can accept undefined as a value empty string will show up as an empty string. So we can check it downstream instead

okdistribute: Or should it be able to return undefined?

gmaclennan: Just completed a big overhaul of this code, since the type checking needed to be much more thorough not just on select fields. Pushing now.

okdistribute: Nevermind, removed this. Although this function probably shouldn't return empty string for undefined

index.js
```diff
@@ -221,35 +221,45 @@ function createServers (done) {
...
-      done()
-    }, 1000)
-  })
+  logger.info('Server ready, checking front-end is loaded')
+  // If the window is still loading, wait for it to finish before continuing
+  // win.webContents.isLoading() does not seem to work here on Windows
+  if (!win._didFinishLoad) {
+    logger.info('Front-end still loading, check again once loaded')
+    win.webContents.once('did-finish-load', () => {
+      win._didFinishLoad = true
```
okdistribute, line 236: This is not critical in this case, but let's not get in the habit of adding properties to objects where the lifecycle is managed by upstream modules (in this case electron window)

gmaclennan: Yes, very much agreed. We need a separate state management for each process (starting, started, closing, closed, errored). This will serve as a hacky short-term fix.

okdistribute: Sure. Something to watch for if we update electron.

PR #396, opened 2020-08-17
fix: Notarize Mapeo for MacOS — fix warning / unable to open app

gmaclennan:
Since MacOs 10.15 when a user downloads and tries to open Mapeo, they get an error that it cannot be open due to Apple not being able to check for malicious software:

image

This PR notarizes Mapeo using Apple's server by following this guide: https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/

Tested locally, uploading and re-downloading Mapeo to trigger this, it does seem to work.


gmaclennan:
@okdistribute I have downloaded the generated app and it works on my machine without issues, so I think this is good to merge. Go ahead and merge if you can't think of any other issues.

okdistribute:
Thanks @gmaclennan ! It looks like you added `build.js.map` file, but that doesn't seem to be related to this PR

PR #401, opened 2020-08-19
fix: Mapeo config directory should be removable

okdistribute:
You'd get `EPERM` permissions errors if you tried to remove the Mapeo config directory using `rm -rf ~/.config/Mapeo`

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ const mkdirp = require('mkdirp')
 const series = require('run-series')
 const styles = require('mapeo-styles')
 const rabbit = require('electron-rabbit')
+const chmod = require('chela').mod
 
 const app = electron.app
 const BrowserWindow = electron.BrowserWindow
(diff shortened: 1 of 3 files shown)


gmaclennan (commented):
Setting permissions to 0777 without user knowledge is not ideal: this folder can contain the sensitive project key. Also best not to make files executable (only directories) and I think `chela` is recursive. Can we modify this to only recursively set directories to u+rwx and files to u+rw and leave group and world permissions the same?

okdistribute:
@gmaclennan why do directories need to be executable?

gmaclennan:
Might be a MacOS thing? But also found this https://unix.stackexchange.com/questions/21251/execute-vs-read-bit-how-do-directory-permissions-in-linux-work

okdistribute:
@gmaclennan how critical is it that files are not executable by the user? It would add a significant amount of code and time right now when I have some other things to work on..

gmaclennan:
I think it's fine. It's a small security risk (it introduces a vector for introducing executable code), but not significant I think.

okdistribute:
Opened an issue for this

PR #402, opened 2020-08-20
fix: Only save the language to config when changed manually. Fixes #400

### LINKED ISSUE #400: Mapeo 5.3.1 stays in English when Mac System language is in Spanish
**Describe the bug**
If the system language is set as Spanish, on a mac, (restarted mac with Spanish set), and you open Mapeo, it all appears still in English.
For reference, I looked back at some earlier versions, and with v5.0.0 it appears as expected in Spanish; in v5.1.0 it already isn't working in Spanish, and it was noted during testing of 5.3.1.
I don't know if this is a bug on windows machines as well as macs.

@jencastrodoesstuff  experienced the same problem

**Expected behavior**
If language of system changes then language of Mapeo should change. Definitely to Spanish anyway as we have it translated.

**Desktop:**
 - OS: ios
 - Version: 10.14.6
 
**Mapeo Version:**
 - Desktop [e.g., 5.3.1]

okdistribute:
### Contributor checklist:

- [NA] If my changes depend upon an update to a dependency, I have updated the package-lock.json file using `npm install --package-lock`

### Description

Mapeo was saving the system locale to the config always on startup, so if you change system languages and reopened Mapeo it wouldn't also change system languages.

This now only persists the change in language if the user explicitly requests to change the language through the User Interface.

### DIFF
--- a/src/main/i18n.js
+++ b/src/main/i18n.js
@@ -9,12 +9,17 @@ const translations = {
   pt: require('../../messages/main/pt.json')
 }
 
+// We only support generalized locales for now (i.e., no difference between
+// Spanish/Espana and Spanish/Latin America)
+function getSystemLocale () { return app.getLocale().substr(0, 2) }
+
 // defaultLocale is the default local of the app, not the user's locale.
 class I18n extends EventEmitter {
   constructor (defaultLocale = 'en') {
     super()
     this.locale = defaultLocale
     this.defaultLocale = defaultLocale
+    logger.info('Locale', this.locale)
     this.t = this.formatMessage.bind(this)
   }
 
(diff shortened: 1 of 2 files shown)


gmaclennan (commented):
How would this behave if the system locale is a language that Mapeo does not yet support? I know in the front-end, using react-intl, I think it will fall back to the default messages (in English) but I am not sure about the back-end?
On mobile there is a check for supported locales a [fallback to `en`](https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/context/IntlContext.js#L53-L68) but not sure if this is needed here. (FYI on mobile we also support dynamic language changes when the system language changes, but I think on Desktop the expectation is that if the user switches the system language, then the app must be restarted to update the locale. This is the case on Mac at least).

okdistribute:
@gmaclennan the fallback is English for both backend and frontend, this already works in `formatMessage` line 27

gmaclennan:
👍 LGTM, although maybe should check checks pass first?

Issue #410, opened 2020-09-01
Save Report  -  button not saving PDF & wrong label


jencastrodoesstuff:
**Doesn't print or save a PDF**
A Button is no longer a print function but a save PDF function - Update label.
B. Save Prompt opens but does not save a pdf. 
C. No message indicates success or report last created date.

**Can you do the same thing twice?**
The exact steps to see the problem again.
1. From the reports view click "Print" (it doesn't mater what the filter state is.)

2. Mac save dialoge opens. Specify location. Name file, Click Save

3. go to specified location too locate saved PDF report 
4. There is no report

**Expected behavior**
A -Save PDF button says what it does
B - that a PDF is actually saved
C - After save, a message confirming save time and a location

To find the PDF in the specified location
.

**Screenshots**
<img width="1286" alt="Screen Shot 2020-08-31 at 9 25 53 PM" src="https://user-images.githubusercontent.com/33264095/91785405-567ed180-ebd3-11ea-9318-f32c11632095.png">

<img width="1072" alt="Screen Shot 2020-08-31 at 9 26 17 PM" src="https://user-images.githubusercontent.com/33264095/91785430-60a0d000-ebd3-11ea-88ff-d928bb959819.png">

<img width="903" alt="Screen Shot 2020-08-31 at 9 26 58 PM" src="https://user-images.githubusercontent.com/33264095/91785445-672f4780-ebd3-11ea-8556-dac21981e3de.png">

**Desktop:**
 - OS: iOS
 - Version: iOS 10.15.6

 
**Mapeo Version:**
 - Desktop: Md v5.4.0.beta7

okdistribute:
![Screenshot from 2020-09-15 14-47-06](https://user-images.githubusercontent.com/633012/93268689-77622d80-f762-11ea-9140-910862b7c63e.png)

jencastrodoesstuff:
@okdistribute  fascinating that your screen capture says save and mine says print.
v5.4.0-beta9
<img width="603" alt="Screen Shot 2020-10-08 at 2 33 58 PM" src="https://user-images.githubusercontent.com/33264095/95499485-552f8a00-0973-11eb-9784-38ec7d7da8a0.png">

Issue #409, opened 2020-08-31
Point in each Report Observation page


okdistribute:
**Describe the idea**
For each page in the report, there is a minimap (miniature map) of the location where the observation took place. There should be a little point or icon in the middle of the map to show where the observation is.

**Have you thought about what it might look like?**
I already wrote a short exploration of this, but the pngjs library is not loading properly in the background window on production mode. I had to remove this functionality in https://github.com/digidem/mapeo-desktop/commit/3c6dad02a332eec38f3d192d86de0db29317b566

PR #411, opened 2020-09-02
fix: Prepare PDF reports for release 

gmaclennan:
### First time contributor checklist:

- [x] I have read the [README](https://github.com/digidem/mapeo-desktop/blob/master/README.md) 

### Contributor checklist:

### Description

Tracking fixes needed for Reports prior to release:

- [x] Fix embedded map size
- [x] Add marker for observation in embedded map
- [x] Show portrait images correctly
- [x] Allow saving reports as PDF
- [x] add warning and feedback for saving large reports 
- [x] Show maps using same style as main MapFilter map
- [x] Only define `mapboxAccessToken` once (pass in URL to backend)
- [x] Don't show "Detalles" header if there are no fields for an observation
- [x] Translate report field names ("Detalles", "Registrado", "Ubicación")
- [x] Add concurrency to static maps to speed up reports
- [x] Cache static maps in the browser
- [x] Support different field types (number, select, multi-select) in report
- [x] Use custom map style in report (separate PR: https://github.com/digidem/mapeo-desktop/pull/423)

**Features to be added later (after v5.4 release):**

- Add report printing
- Add overview map of all observations
- Add cover page with summary of all observations
- Allow configuration of fields on cover page
- Show preview of entire report and total pages

I moved some code around in 05a7098 to avoid potential bugs (using prevFilter to work around a bug in @react-pdf).

**Report printing**

I tried rendering the report into an iframe and calling print on that, but after the print window opens and the user clicks print it just hangs. We can try `<webview>` or opening a new BrowserWindow with the PDF as the source. The print button also needs to print the entire report (rather than the first 5 pages).

**Report saving**

The print button currently works as a save button, but only saves the first 5 observations. Need to check with @jencastrodoesstuff about what the default name of the saved report should be (currently `report.pdf`). Could create the filename from the filters (e.g. start and end date).

**Outstanding bugs**

There is a bug where the report component will stop rendering if you quickly click the filters a few times. This seems to be caused by `react-pdf` **viewer** trying to setState after the component is unmounted. I _think_ this happens when the user changes the filter in the exact moment between the pdf blob URL being ready but before the pdf viewer calls `onLoadSuccess`.


okdistribute:
I tried to add a marker in https://github.com/digidem/mapeo-desktop/issues/409, ran into a bug with the pngjs package running in a background winder

patches/@react-pdf+renderer+2.0.0-beta.14.patch
```diff
@@ -0,0 +1,136 @@
...
+-    this.instance.updateContainer(this.props.document);
++  _proto.componentDidUpdate = function componentDidUpdate(prevProps) {
++    if (prevProps.document !== this.props.document) {
++      this.instance.updateContainer(this.props.document);
++    }
+   };
+ 
+   _proto.componentWillUnmount = function componentWillUnmount() {
+     this.renderQueue.end();
++
```
okdistribute, line 49: This file makes me more nervous the larger it gets as we may diverge in months time. Can we have a fork of react-pdf/renderer and open a pull request?

src/renderer/components/MapFilter/ReportView/ReportView.js
```diff
@@ -1,247 +1,45 @@
...
   onUpdateObservation,
   onDeleteObservation,
   mapboxAccessToken,
-  bounds,
   mapStyle,
   presets,
   filter,
   getMediaUrl,
   ...otherProps
 }: Props) => {
```
okdistribute, line 22: The reason I kept this stuff inside this file is that it's difficult to know which file contains which logic just by their names -- ReportView, ReportViewContent, PDFReport, PdfViewer?

Is there a way we can make these names more memorable?

gmaclennan:
Hey I think the mapbox access token should be passed as a URL param. I think this will make things simpler in the future (e.g. if we allow the user to use their own access token). Did you see https://github.com/digidem/mapeo-desktop/pull/303#issuecomment-682891903

okdistribute:
@gmaclennan can we implement that later? I think that might be premature complexifying for a feature we don't have yet

src/renderer/components/MapFilter/ReportView/ReportView.js
gmaclennan: Yes I agree (naming things is hard!). I tried to follow this convention for all the views:

`ComponentNameView` e.g. `MediaView` = just wraps the component in `ViewWrapper` which passes the filtered observations and accessor functions, doesn't render anything

`ComponentNameViewContent` e.g. `MediaViewContent` is the logic of the view.

Anything else is sub-components that make sense to move into their own file.

Happy for suggestions for clearer naming.

gmaclennan:
Yes no worries about implementing later.

okdistribute:
I just realized that printing lots of observations might take a while because mapbox-gl needs to be run serialized, so we might want to parallelize the mapbox-gl instance... that can be done in a separate PR as it's an optimization, but I think it would be ideal to have this before release as people might naively try to save 300 observations as a PDF. This makes me think, should we bring back the print dialogue to let the user know how many pages it'll end up being and how long it might take to render?

We might need to UX burrow on this.

patches/@react-pdf+renderer+2.0.0-beta.14.patch
gmaclennan: Yes I made PRs today https://github.com/diegomura/react-pdf/pull/995 and https://github.com/diegomura/react-pdf/pull/996.

Forking would be a pain because it would mean publishing all the linked packages from the v2 monorepo. I think the .patch is ok, but if the PRs are not merged in the next release we should re-visit.

okdistribute: We can depend upon github forks so it doesn't have to be published as a fork; but sure, this is fine and hopefully they'll merge and publish.

gmaclennan:
Yeah, good point. We could create multiple instances of hidden-mapbox to parallize things. As for larger files, I think we can make an estimate of the number of pages based on number of observations. I think maybe an alert dialog when the user clicks save/print when the report is > 50 observations, and say something like that "This will take a while, are you sure?\n\nYour report is more than XX pages long, this will take several minutes to prepare, are you sure you want to save/print?"

patches/@react-pdf+renderer+2.0.0-beta.14.patch
gmaclennan: Yeah unfortunately most front-end packages don't work from github forks because what is published to npm is built code. Node coding is much easier in this respect! Hence why I lean on patch-package for front-end stuff as the least-friction option. For mapeo-mobile I mantain a [README](https://github.com/digidem/mapeo-mobile/blob/develop/patches/README.md) in the patches folder to track what the changes are in there for. I have found it easier to maintain than publishing forks on npm.

gmaclennan:
@okdistribute added a quick fix for markers for observations. Could probably do with better styling, but we can do that later. Sorry I realised that I formatted the file (everything under `MapFilter` when I merged was formatted with Prettier. In mapeo-mobile I run prettier as a git hook, it makes writing React JSX so much easier. I didn't want to mess with the rest of the code and formatting though). My editor runs prettier automatically if it is configured for a repo, and I forgot to save without formatting for this commit and I'm about to run for dinner, so wanted to share anyway.

okdistribute:
> Yeah, good point. We could create multiple instances of hidden-mapbox to parallize things. As for larger files, I think we can make an estimate of the number of pages based on number of observations. I think maybe an alert dialog when the user clicks save/print when the report is > 50 observations, and say something like that "This will take a while, are you sure?\n\nYour report is more than XX pages long, this will take several minutes to prepare, are you sure you want to save/print?"

@gmaclennan regarding printing, @jencastrodoesstuff says that users send the pdf as an attachment to an e-mail rather than printing directly from Mapeo. This means that having Save to PDF is good enough for release to get something they can use while we figure out how printing can work seamlessly. I agree that we should include some warning, like you mention!

patches/@react-pdf+renderer+2.0.0-beta.14.patch
okdistribute: ah thanks for that, that's helpful!

src/renderer/components/MapFilter/ReportView/ReportView.js
okdistribute: I'd put the ViewWrapper code in `index.js` but since this is the case in other components as well, maybe too much work for what it's worth at this moment.. label it under 'future developer experience' issues

gmaclennan:
Great yes if that's the case for Jen then I think we're ok to push printing. It might frustrate some other users though, but maybe ok since they can print via a pdf reader (if they have one installed)

src/renderer/components/MapFilter/ReportView/PDFReport.js
```diff
@@ -103,36 +101,53 @@ const PDFReport = ({
...
+          : null
+        }
       </View>
+      <ObservationRHS observationView={view} />
+    </View>
+  )
+}
+
+function fields (view: ObservationView) {
+  return (
```
okdistribute, line 149: this fixes the bug where Detalles would be shown no matter what, even if no fields.

We need to make this code better tho! This was a quick hack

gmaclennan:
@okdistribute made a few fixes today, adding concurrency and caching, as well as map styling and fixing up styling a little. I noticed that we do not yet manage field display in the same way as the ObservationDialog.

Concurrency does not actually seem to make much difference. I measured render of a PDF with 100 observations, and with concurrency of 1 and concurrency of 10 it is all within 12000ms +/- 500 (tested the prod build with `npm start`). I think this is because most of the time taken is fetching map resources, and most of the heavy lifting by mapbox gl is done on workers, which I think are shared by all mapbox instances, so we don't actually get a speed up. The caching makes the bigger difference when switching between different filters. We should start adding cache headers to images too.

I am seeing a strange issue cropping up which I can replicate with concurrency 1: when I switch to a local style then sometimes the PDF fails to load. Inspecting the background process I see that it is being held up by the request to http://localhost:5000/ for the style.json. It sometimes does seem to work, but sometimes just hangs (request `pending` in the network inspector). I'm not sure if this is an issue with `ecstatic` or an issue with electron and making a request to a server running in the render process, but I think this is potentially a blocker for reports.

@jencastrodoesstuff a question for PDF reports: We have two groups of fields:

1. **Details**: These are the fields that are defined in the presets.
2. **Additional Details**: These are additional fields defined on the observation which are not defined in the preset. These exist because a field might have been added with an old preset definition, and that preset was changed.

Additional Details do not show at all on mobile. On Desktop they show up under the collapsible section "Additional Details" when you edit an observation.

How do you think these additional details should show up on PDF reports? Should they be shown at all? Should they appear in a separate section?

jencastrodoesstuff:
Field display on forms

All fields should have the capacity to be displayed but I believe it would be ideal for the user to control this using the toggles.
That being said,the default toggles positions for additional feilds should be off.
Screen Shot 2020-09-03 at 10 45 04 AM

jencastrodoesstuff:
Re: field display, currently we do not display multi-select and select fields in a way that uses the label. Do you think we should block the release until this is fixed?

It is not a block. It just makes writing configs difficult.

okdistribute:
@gmaclennan can you open an issue for your bug you mentioned about 'siwtching to a local style?' I'm not exactly sure what you mean by that and how to reproduce.

okdistribute:
@jencastrodoesstuff I'm going to open a new issue for this, thanks for reporting..

okdistribute:
https://github.com/digidem/mapeo-desktop/issues/415

okdistribute:
I created a project to track related blocker bugs https://github.com/digidem/mapeo-desktop/projects/6

gmaclennan:
> @gmaclennan can you open an issue for your bug you mentioned about 'siwtching to a local style?' I'm not exactly sure what you mean by that and how to reproduce. Also I want to avoid this long pull request with 5 conversations/bugs happening at the same time, if we could open issues that would make our lives long-term easier I think

https://github.com/digidem/mapeo-desktop/issues/416

okdistribute:
I think this PR looks good and we can do a last review & merge it. I'd then like to open an issue for saving reports as a PDF. All other issues have been documented, including label/value bugs in reports. I'm not sure about the last one though regarding select/multiselect fields in reports; what is the actual issue exactly? perhaps we can open another issue for that as well

gmaclennan:
Would it make more sense to know we have reports working without blocking bugs before merging? I feel like it's useful to have everything in one place and get it ready so we can check whether some of the code here is responsible for bugs before we merge it in and loose track.

okdistribute:
I implemented offline map style rendering [in a separate PR](https://github.com/digidem/mapeo-desktop/pull/423) because it is a significant piece of code and is a bug we can isolate there.

okdistribute:
This PR fixes #416, #412, #410, #409

okdistribute:
@gmaclennan This branch is ready for merge.

Would it be better to do a merge commit for this branch? Our project settings force a squash & merge through GitHub, but this branch includes many different features and bug fixes that are separated into small chunks. I think it would be ideal for future team's sake to be able to understand which commits go together.

gmaclennan:
I'm ok with this being a merge commit in this case, since the feature of reports is split across two PRs.

Once we have more comprehensive tests I think it makes more sense to strictly enforce squash and merge, so that all commits to master pass tests. This will mean that large new features are squashed to a single giant commit, but the history remains in the PR, and the body of the squashed commit refers to the individual commit shas.

PR #420, opened 2020-09-08
chore: move flow types into mapeo-schema

gmaclennan:
`mapeo-schema` used to generate flow types from JSON schema. Mapeo Desktop defined more strict types manually. `mapeo-schema@2` includes those manually defined types, and this PR switches Mapeo Desktop to use the types from mapeo-schema.


okdistribute:
LGTM

PR #437, opened 2020-10-07
fix: Fallback to exporting preview-sized images if originals are missing

gmaclennan:
### Contributor checklist:

- [ ] I have walked through the [QA Manual Testing
- [x] My changes are ready to be shipped to users on Windows, Mac, and Linux using the production version of the application built with electron-builder.

### Description

It is possible for Mapeo Desktop to be missing originals of images attached to observations. This is not obvious to the user in the app because we always show the preview-sized images (1,200px). This can happen because phone-to-phone sync only syncs thumbnails and previews, not originals, to reduce disk space usage on phones. As a result if a phone has never synced with Desktop, only the previews and thumbnails will be available.

Currently, if the user tries to export observations that link to images with missing originals, via either the Webmaps export, or the data export with "Originals" selected, then the images will be missing from the exported data, with no warning shown to the user other than a file `missing.txt` included in the exported Zip. There is no way for the user to export a webmap with the images available (the preview-sized images).

This PR changes export behaviour to fallback to including preview-sized images if no originals are available. It changes the text files included in the Zip export to `Export Errors.txt`, which lists images that are not included in the export, and `Missing Originals.txt` which lists the images for which originals are not available.

The goal of this PR is to ensure that the user has a way to export and access the images that are available, even if the originals are not. The preview size (1,200px) is adequate for most online viewing and sending electronically.

The current method of feedback to the user about what has happened, via a text file included in the export, is not ideal, and at a later date we should add clearer feedback to the user with clear actions they can take to fix this.

For WebMap exports, it may make sense to default to preview-sized images to keep the webmap bundle size down anyway.

I have tested this by doing the following exports from the app with test data:

1. Export webmap with all images present.
2. Export webmap with missing originals but preview sizes present.
3. Export webmap with missing images (originals and previews)
4. Export data zip with "originals" selected, but originals missing (previews are included in export, `Missing Originals.txt` is in export with explanation).
5. Export data zip with "previews" selected and all previews available (previews are included, as expected).
6. Export data zip with "originals" selected, but missing images (no images in export, `Export Errors.txt` shows explanation).

### DIFF
--- a/src/renderer/create-zip.js
+++ b/src/renderer/create-zip.js
@@ -5,62 +5,105 @@ import once from 'once'
 import logger from '../logger'
 
 const concurrency = 3
+const errorsMsg =
+  'There was an error trying to export these files, so they are missing from this export:'
+const missingOriginalsMsg =
+  'The original size of these files could not be found, only preview size ' +
+  '(1,200 pixel) images are included. This can happen because the phone that ' +
+  'took the photos has only synced to other phones, and not directly to ' +
+  'Mapeo Desktop. To try fixing this, find the phone that took the photos ' +
+  'and sync it with Mapeo Desktop before exporting again.'
 
 /**
  * Create a zipfile from a collection of strings/buffers and remote files
  * (remote files will be downloaded concurrently and streamed to the zipfile)
  *
  * @param {Array} localFiles Array of files to add to the zip archive. Must have
- * properties `data` which should be a Buffer or String, and `name`
+ * properties `data` which should be a Buffer or String, and `metadataPath`
  * which is the path to the file in the zipfile. Can also include options from
  * https://github.com/thejoshwolfe/yazl#addfilerealpath-metadatapath-options
  * @param {Array} remoteFiles Array of files to add to the zip archive. Must
- * have properties `url` and `name` which is the path to the file in the
- * zipfile. Can also include options from
+ * have properties `url` and `metadataPath` which is the path to the file in the
+ * zipfile. Can also include `fallbackUrl` which will be tried if `url` fails
+ * with an error, and any options from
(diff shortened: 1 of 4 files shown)


okdistribute:
This removes the `missing.txt` file - is there any documentation that references that file which needs to be changed?

Does this need to make it into 5.4?

gmaclennan:
> This removes the `missing.txt` file - is there any documentation that references that file which needs to be changed?

I'm not aware of this being documented anywhere. I originally included this as a kind of "self documentation" for any user dealing with exported data, as a short-term step before we have a better UX in the app to show this kind of export issue.

> Is this for 5.4 release?

I don't think it's urgent. It came up in a convo with Arky because workshop participants had been trying to use the "Export to webmap" functionality and it just didn't work. I felt like it's worth fixing since the webmaps functionality can be useful for some users, and since we have the service mainly working it's good to get this fixed.

okdistribute:
@gmaclennan sorry for not being able to parse your responss, but do you want this in 5.4.0 release?

gmaclennan:
Sorry about the lack of clarity. If this does not involve delaying the v5.4 release, then yes, include it. If including it would mean delaying the release, I think it is ok to leave until the next release, because it is not urgently needed.

okdistribute:
It looks good to me! We can get it in 5.4

src/renderer/create-zip.js
```diff
@@ -5,62 +5,105 @@ import once from 'once'
 import logger from '../logger'
 
 const concurrency = 3
+const errorsMsg =
+  'There was an error trying to export these files, so they are missing from this export:'
+const missingOriginalsMsg =
```
okdistribute, line 10: This could be internationalized

gmaclennan: Done

PR #444, opened 2020-10-12
feat: Help -> Open User Data Folder menu button

okdistribute:
Is this menu button useful to folks?

Screenshot from 2020-10-12 15-34-47
Screenshot from 2020-10-12 15-35-04

### DIFF
--- a/messages/main/en.json
+++ b/messages/main/en.json
@@ -51,6 +51,7 @@
   "menu-selectall": "Select all",
   "menu-debugging": "Debug",
   "menu-report": "Report issue...",
+  "menu-open-user-data": "Open User Data Folder",
   "menu-check-for-updates": "Check for updates...",
   "menu-get-beta": "Test the latest beta version (unstable)...",
   "menu-downgrade-beta": "Go back to latest stable version...",
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -70,7 +70,7 @@ function menuTemplate (ipc) {
           click: async function (item, focusedWindow) {
             const result = await dialog.showOpenDialog(
               {
-                title: t('menu-import-configuration-dialog'),
+                title: t('menu-import-configuration-title'),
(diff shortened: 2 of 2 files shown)


gmaclennan:
Yes, this could be very useful, particularly on Windows where opening this folder is tricky to explain.

However, this is also a risk, since accidentally moving files in this folder can break things. We do not want the "average user" clicking around and exploring and changing things without ensuring that they first understand what the consequences might be.

I think we should aim for users never needing to touch this folder. However, short-term it seems like users do need to access this folder for debugging, so short-term it is useful to have a way for users to access it.

One small tweak might be to add both this and the "Beta channel" option to a sub-menu called "Advanced…" under the Help menu. I think the "Help" is the most obvious location for this menu option, because the user should only be accessing this as part of a trouble-shooting workflow.

aliya-ryan:
Thanks Karissa - I agree with Gregor that this is very useful, also on a mac it is hard to explain how to find this hidden folder, but similarly I would want to be very cautious about risks of people accessing it and doing things in it without knowing that that is where all the data is stored and can be lost or broken from. 

I wonder if before opening the folder it could first go to a popup box that gave a warning to people so they would have to click an additional 'OK' to understanding risks of going into that folder - or some other UX feature to safeguard against data loss or corruption. 

I would be hesitant to say to implement it right away without any further precaution or information to the user.

gmaclennan:
Ok, let's hold off merging this until:

1. We move it under Help --> Advanced…
2. We add a warning dialog that appears when the user selects this menu item and explains what this folder is and the dangers of moving / deleting files / folders in it, with two actions: (A) Continue to open folder; (B) Cancel / go back.

gmaclennan (changes requested):
Changes requested in https://github.com/digidem/mapeo-desktop/pull/444#issuecomment-708340265

PR #456, opened 2020-11-24
fix: PDF reports

gmaclennan:
Work in Progress to fix the new PDF reporting feature to meet user requirements for release:

1. Allow users to navigate through all pages of a report (currently only shows the first ~5 pages)
2. Remember page position when navigating away and returning
3. Show how large the report is (how many observations are in the report)
4. User can edit the observation currently displayed in the report preview
5. User can view report preview in languages with non-latin characters (Thai)

<ins>Tasks:</ins>

- [x] Render pages on demand, allowing user to navigate through all pages
- [x] Cache previously rendered pages
- [x] Eagerly render next page before the user navigates (reduce loading delay on navigation)
- [x] Add new page navigator design
- [x] Adjust report preview design so page edges are clear
- [x] Improve loading view so that layout does not jump around
- [x] Allow vertical scroll
- [x] Add Edit button (brings up edit observation dialog)
- [x] Change icon on Save button
- [x] Display # of observations in report and total observations in Mapeo
- [x] Fix display of non-Latin characters
- [x] Add page numbers to report footer (stretch goal)
- [x] Implement new report layout & design from Aldo (stretch goal)
- [x] Show icons for each category (stretch goal)
- [ ] ~~Cache and use page renders for final render of page to speed up export (stretch goal)~~
- [ ] ~~Show progress during rendering of whole report on save (stretch goal)~~
- [ ] ~~Reduce report preview width to screen width on smaller displays~~

[Storybook preview of report view](https://deploy-preview-456--mapeo-desktop-design.netlify.app/?path=/story/reportview--all) (Requires Google Chrome)


gmaclennan:
### Known issues

> * [ ]  Reduce report preview width to screen width on smaller displays

Current behaviour: If the users screen is small such that the width of the report is larger than their screen, then a horizontal scroll bar will show.

> * [ ]  Cache and use page renders for final render of page to speed up export (stretch goal)
> * [ ]  Show progress during rendering of whole report on save (stretch goal)

Current behaviour: During export a dialog will appear showing a spinning progress animation, but there is no indication about how long it will take. The progress animation can get stuck.

image

This is a hard fix: it's hard to calculate how long the report will take to render, it's hard to get updates during rendering (it's a single task that we start and wait for it to finish) and because it takes a lot of processing power, it freezes the loading animation. There are work-arounds for all of this, but it is quite a bit of work to fix.

* Occasional bug where report gets stuck

It seems like if the user clicks around lots, back and forth to reports, then there is a very rare case where the report will just never appear, and Mapeo needs to be restarted. I have done a few hours of bug-hunting and am pretty sure this is a bug in the underlying PDF library, and it is very hard to reproduce. @jencastrodoesstuff do you think we can work around this for now?

### Still TODO

* [ ] Disabled state for page navigations buttons

The page navigation buttons (for next page and previous page) are disabled if the user is on the first page or the last page (e.g. if the user clicks then it does nothing). However, we do not have a design for what a disabled button should look like. Normally a disabled button is grayed out (see https://material.io/design/interaction/states.html#disabled) but I was not sure how to design these buttons. @macandrow can you suggest a design for disabled navigation buttons?

* [ ] Empty fields in report

In testing it seems like report pages seem to show all fields for a preset (or maybe all fields that appear across all data in Mapeo?) and for each shows "No Answer". I think we should maybe not display fields that are not filled in, only showing completed fields? What do you think @jencastrodoesstuff ?

image

* [ ] Report maintains page position when navigating away and back

Currently if you navigate away (e.g. to map or media view) and back, then the report preview will return to page 1. I know @jencastrodoesstuff identified this as an issue before, but does it remain an issue now that you can edit the observation from the report view?

jencastrodoesstuff:
the animation bug is fine for now.

jencastrodoesstuff:
Agree about hiding empty fields.

jencastrodoesstuff:
switching views resetting to page 1 is ok for now.  Yes with edit available from report pages this is less of a workflow issue.

PR #457, opened 2020-11-25
Add MapeoWeb Sync

RangerMauve:
### First time contributor checklist:

- [ ] I have read the [README](https://github.com/digidem/mapeo-desktop/blob/master/README.md) 

### Contributor checklist:

- [ ] My contribution is **not** related to translations. _Please submit translation changes via our [Mapeo Crowdin project](https://crowdin.com/project/mapeo-desktop)_
- [ ] My commits are in nice logical chunks with [good commit messages](http://chris.beams.io/posts/git-commit/)
- [ ] I have walked through the [QA Manual Testing
- [ ] If my changes depend upon an update to a dependency, I have updated the package-lock.json file using `npm install --package-lock`

### DIFF
--- a/src/background/mapeo-core/mapeo.js
+++ b/src/background/mapeo-core/mapeo.js
@@ -135,7 +135,19 @@ class MapeoRPC {
     })
   }
 
-  syncStart (target = {}, createFile) {
+  syncConnect(url) {
+    try {
+      logger.debug(
+        'Connecting to MapeoWeb',
+        this.encryptionKey && this.encryptionKey.slice(0, 4)
+      )
+      this.core.sync.connectWebsocket(url, this.encryptionKey)
+    } catch (e) {
+      this._handleError('syncConnect', e)
+    }
+  }
+
+  syncStart (target = {}) {
     logger.info('Sync start request:', target)
 
     const sync = this.core.sync.replicate(target, {
--- a/src/background/mapeo-core/mapeo.js
+++ b/src/background/mapeo-core/mapeo.js
@@ -135,7 +135,7 @@ class MapeoRPC {
     })
   }
 
-  syncConnect(url) {
+  connectCloud(url) {
     try {
       logger.debug(
         'Connecting to MapeoWeb',
@@ -144,7 +144,7 @@ class MapeoRPC {
       )
       this.core.sync.connectWebsocket(url, this.encryptionKey)
     } catch (e) {
-      this._handleError('syncConnect', e)
+      this._handleError('connectCloud', e)
     }
   }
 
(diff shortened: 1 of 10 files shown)


gmaclennan:
So I think the process to merge this is for someone on the programs team to find some time to do some testing of a build from this branch, so that we can check that basic functionality is working as expected. Since we are QA testing `master` right now for the v5.4 release, I don't want to merge anything new in until we know it works well.

RangerMauve:
@gmaclennan How do you feel about the API method name I added? I'm not sure I'm 100% into it. 😅

src/renderer/components/SyncView/index.js
```diff
@@ -39,8 +40,12 @@ const fileDialogFilters = [
 
 const SyncView = ({ focusState }) => {
   const cx = useStyles()
+  const {
+      metadata: { syncServer }
+  } = useConfig()
   const listenForSyncPeers = focusState === 'focused'
-  const [peers, syncPeer] = usePeers(listenForSyncPeers)
+  const [peers, syncPeer, connectCloud] = usePeers(listenForSyncPeers)
```
gmaclennan, line 47: There are 3 states here:

1. config is loading
2. config is loaded
3. error loading config

Right now it's not clear what happens during the "loading" state (e.g. when syncServer and Encryption key are undefined because the async call has not completed). We should be explicit about the loading state, and render a loading indicator. We should also catch any error with loading the encryption key and metadata. What should we show to the user if there is an error loading metadata or encryption key (which could be due to a future change in the config file format for example, or a corrupt config file).

gmaclennan (commented):
Ok, generally looking good here. Left a couple of comments

src/background/mapeo-core/mapeo.js
```diff
@@ -130,6 +130,19 @@ class MapeoRPC {
...
   }
 
+  connectCloud(url) {
+    try {
+      logger.debug(
+        'Connecting to MapeoWeb',
+        url,
+        this.encryptionKey && this.encryptionKey.slice(0, 4)
+      )
+      this.core.sync.connectWebsocket(url, this.encryptionKey)
```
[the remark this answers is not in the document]
RangerMauve: The default key gets used by the mapeo-core API.

src/renderer/components/SyncView/index.js
RangerMauve: When syncServer is undefined, the button to connect to the cloud is hidden. Does that help?

Does the config loading error handling need to happen for this PR, or is that something that could be handled after? Feels like the code for that already existed.

PR #462, opened 2020-12-17
feat: Add export for WCMC ICCAs

louh:
### Description

Here's the ICCA export package feature, which covers this part of the workflow:

> User exports the ICCA area, and fills in a separate form as part of the export process answering
questions about the submission and sharing of data with UNEP-WCMC. Data is exported as a
Zip file containing a GeoJSON representation of the area with the attached metadata, and a
separate JSON file of metadata related to the export.

- On the "Territory" tab, clicking on the menu on the upper right of the window reveals a "Export ICCA package" menu item.
  Screen Shot 2020-12-17 at 5 40 49 PM

- As a part of the export process, the user answers a few additional questions.
  Screen Shot 2020-12-17 at 5 40 58 PM

- User is notified of required form fields if they are missing.
  Screen Shot 2020-12-17 at 5 41 07 PM

- The file is saved using the file system dialog. It borrows from the same `.mapeomap` process used to export a webmap. The file is in reality a zip file, with the `.mapeomap` extension:
  Screen Shot 2020-12-17 at 5 44 15 PM

- The five questions in `metadata.json` are output like so:
  ````json
  {
    "communityConsent": "true",
    "makePublic": "true",
    "communityOriginalName": "foo",
    "communityEnglishName": "",
    "relationship": "Representative or associate of the community/indigenous people",
    "contact": "bar"
  }
  ````

#### Questions

- How do we make the "Export ICCA package" menu item only display if the ICCA plugin is present?
- Is this the desired file export name (with .mapeomap extension)?
- I took a stab at the structure of `metadata.json`, anything else I should be aware of for this output?

#### TODO / next steps
- I haven't looked into the app's testing infrastructure yet; if things are going in the right direction, integration tests will be next on my to-do list (to the extent there's already a framework for this set up)
- import process into the webapp!


gmaclennan:
> #### Questions
> * How do we make the "Export ICCA package" menu item only display if the ICCA plugin is present?

We can evaluate the presets in the current configuration, and show it if there is a preset with the tag `protection_title=icca`:

```js
const isIccaConfig = Object.keys(presets.presets).find(key => presets.presets[key].tags.protection_title === 'icca')
```

You can get the current presets via IPC call as in https://github.com/digidem/mapeo-desktop/blob/master/src/renderer/components/MapEditor/index.js#L249

> * Is this the desired file export name (with .mapeomap extension)?

Let's use `.mapeoicca` to keep this separate and avoid accidental uploading of the wrong file.

> * I took a stab at the structure of `metadata.json`, anything else I should be aware of for this output?

Looks good to me.

> #### TODO / next steps
> * I haven't looked into the app's testing infrastructure yet; if things are going in the right direction, integration tests will be next on my to-do list (to the extent there's already a framework for this set up)

Unfortunately we do not have strong testing infrastructure set up yet for Mapeo Desktop, and it is on our TODO list. I had been planning to follow the guidelines from react-testing-library, which seems like a good approach to tests for me.

> * import process into the webapp!

I am thinking now that we should set up a separate instance of the webapp for this, since I think this use-case is different enough to separate it from the existing service which is based on sharing maps. I will try to set up an instance to play with this week.

src/renderer/components/MapEditor/ICCAExportDialog.js
```diff
@@ -0,0 +1,494 @@
...
+  const [value4Error, setValue4Error] = useState(false)
+  const [value5Error, setValue5Error] = useState(false)
+  const ref = useRef()
+
+  // Pre-optimize the parent element for the "shake" animation when there's
+  // an error. If we don't do this, the first time the animation plays,
+  // it feels janky
+  useEffect(() => {
+    if (!ref.current) return
+    ref.current.parentNode.style.willChange = 'transform'
```
gmaclennan, line 126: This approach feels like it is fighting against React and is fragile (not clear what parent node will be if we change things). May I suggest that we add a `onFormError` prop to `EditDialogContent`, and a handler in `ICCAExportDialog` that sets a state `shaking=true`, with a timeout to turn it off, then we add the class `shakeX` to the parent `Dialog` component based on the `shaking` state, using the [`classes` property](https://material-ui.com/api/dialog/) exposed by the Material-ui API.

src/renderer/components/MapEditor/ICCAExportDialog.js
```diff
@@ -0,0 +1,494 @@
...
+    animation-duration: 500ms;
+    animation-delay: 0;
+  }
+`)
+
+const EditDialogContent = ({ onClose }) => {
+  const classes = useStyles()
+  const { formatMessage } = useIntl()
+  const [saving, setSaving] = useState()
+  const [error, setError] = useState(false)
```
gmaclennan, line 107: This is ok for now, but I like to avoid boolean states, and instead have a single state for the component as a string, which can be `idle`, `saving`, `error`. This avoids bugs and impossible states if we ever add things to this component (because what the component should show is essentially a combination of all boolean states right now).

src/renderer/components/MapEditor/ICCAExportDialog.js
```diff
@@ -0,0 +1,494 @@
...
+  const [value2, setValue2] = useState('')
+  const [value3a, setValue3a] = useState('')
+  const [value3b, setValue3b] = useState('')
+  const [value4, setValue4] = useState('')
+  const [value5, setValue5] = useState('')
+  const [value1Error, setValue1Error] = useState(false)
+  const [value2Error, setValue2Error] = useState(false)
+  const [value3Error, setValue3Error] = useState(false)
+  const [value4Error, setValue4Error] = useState(false)
+  const [value5Error, setValue5Error] = useState(false)
```
gmaclennan, line 118: Again, ok for now, but this is an indication that we should add a TODO to decide on a form library for Mapeo to cut down on repeated logic like this. I'm thinking either https://final-form.org/react or https://react-hook-form.com

src/renderer/components/MapEditor/ICCAExportDialog.js
```diff
@@ -0,0 +1,494 @@
...
+      setError(true)
+      errorShake()
+      return
+    }
+
+    setSaving(true)
+    const points = await getGeoJson()
+    const metadata = {
+      communityConsent: value1,
+      makePublic: value2,
```
gmaclennan, line 171: Can we write value1 & 2 as boolean values?

gmaclennan (changes requested):
Sorry it took me a while to get to this this week, I've been slow getting started since the holidays.

Can you add an error state for when there is no data in the GeoJSON export (e.g. the user has no ICCA boundaries) with text explaining how the user can resolve this (e.g. draw an ICCA boundary).

I'd like to remove the code that reaches into the DOM to mutate nodes, but the rest looks ok.

src/renderer/components/MapEditor/ICCAExportDialog.js
louh: I went ahead and made this change in the revision because there's going to be a new dialog state (when there's boundary data to export). Not sure if you'd like this, but there are actually two categories of states in this component: the dialog state, and the form state.

The dialog state will have one of three values:
  - `pending` - initial state, checking geo data
  - `idle` - default state, form is displayed
  - `error` - no boundaries to export, show instructions

And there are three form states:
  - `idle` - default state, user is filling out form
  - `error` - a required form field is missing or invalid
  - `saving` - saving is in progress, disable inputs

I didn't want to combine the state values because the form state is really a subset of when the dialog state is `idle` (that is, displaying the form), and I didn't want to overload the meanings (which would make a "form error" state also means "display the form" in _addition_ to "this form has an error state"). Another way to get around this, if you don't like having two states in one component, is we put the form content into its own component.

src/renderer/components/MapEditor/ICCAExportDialog.js
louh: I'll add a TODO for now. I don't have experience with either of these libraries, so I'll await recommendations from the team.

src/renderer/components/MapEditor/ICCAExportDialog.js
louh: Yep. Unfortunately HTML values can only be strings, but we can convert that to boolean values when we write out the JSON.

src/renderer/components/MapEditor/ICCAExportDialog.js
louh: Yep - that's a good call. Thanks for the suggestion.

src/renderer/components/MapEditor/ICCAExportDialog.js
louh: I did have some time to look into these briefly, and here's what I've found - the challenge to overcome is how to forward the form library's props to Material UI components.

Someone wrote a nice tutorial about how [to integrate react-hook-form with Material-UI](https://medium.com/@andresss/using-material-ui-with-react-hook-form-is-this-simple-d8383941fafe). It seems react-hook form exports a wrapper that makes this possible, right out of the box.

It seems that there is not a way to do this with React Final Form, so someone [wrote their own wrapper library](https://github.com/lookfirst/mui-rff) to integrate RFF with Material-UI.

That's an edge toward react-hook-form. I also like the syntax on RHF better as it's just hook-based, so everything works within the component itself. RFF on the other hand is component and context-based, so it's a little more complex to use.

Issue #446, opened 2020-10-14
"Zoom to data" in Territory view doesn't zoom to observations


rudokemper:
**Describe the bug**
It seems that currently, the "Zoom to data" option (View menu) for the `Territory` view is not zooming to points that were collected as Observations using Mapeo Mobile. When you have an instance of Mapeo that has Observations loaded in, but nothing drawn natively in Territory, clicking on "Zoom to data" does nothing.

**Can you do the same thing twice?**
The exact steps to see the problem again.
1. Start with an empty instance of Mapeo Desktop
2. Collect observations on Mapeo Mobile
3. Sync those observations with Mapeo Desktop
4. Ensure that the observations have been synced in the `Observations` view in Mapeo Desktop
5. Navigate to the `Territory` view.
6. Select `View` -> "Zoom to Data".
7. Nothing happens.

**Expected behavior**
I would expect "Zoom to data" to zoom to the maximum extent of all data, both observations as well as data mapped natively in the Territory view.

**Screenshots**
![image](https://user-images.githubusercontent.com/31662219/96002020-aa9de800-0e06-11eb-8f78-c2ac612fded9.png)
This singular data point (collected as Observation) is not zoomed to when doing the above behavior.

**Desktop:**
 - OS: Windows 10 Enterprise
 
**Mapeo Version:**
 - Desktop 5.3.1
 - Mobile 5.0.3.879

gmaclennan:
Fixed in #475

Issue #474, opened 2021-06-08
"Zoom to data" does not always zoom to data


gmaclennan:
In QA testing I found several issues with the zoom to data function. Deciding where to zoom to is not trivial: if it is the bounding box of all the data then if there are points across a wide area, this could end up not "zooming" at all, so in territory view, which does not show data until zoomed in, then it would not show any data. Similarly, zooming to the "center" of the data might not work: imagine 3 points in New York and 3 points in London - the "center" would be in the Atlantic. The current logic sums the points in "buckets" which are ~1.1km square, and considers "data" to be the bucket with the highest number of points, e.g. the "densest" area of data. This is an approximation, but should at least help you find the data. Currently there are several issues with this logic:

- The count in each bucket is stored as a string, e.g. `"1"`, so when adding a new point to the bucket the result was `"1" + 1 = "11"`
- The bucket count does not consider deletions, so you could end up zooming to an area with a large number of deleted points.
- The bucket count does not consider edits - each edit is counted as a new point in a bucket
- The code logic is counting the incorrect points in each bucket, due to async logic within a loop, and variables being hoisted and changed before async code runs.

gmaclennan:
Fixed in #475

PR #475, opened 2021-06-08
fix: Fix "zoom to data": include deletions & correct density calc

gmaclennan:
Fixes most of #474 apart from considering edits. E.g. if you edit a point 3 times, that will be counted as 3 points. This should still work "well enough" to find your data.


gmaclennan:
Also now fixes #446

PR #493, opened 2021-06-25
fix: update electron-builder to support building multiple nsis archs

### LINKED ISSUE #492: Auto-updater does not consistently work with Windows 64-bit and 32-bit builds
Mapeo should support Windows 32-bit machines. To do that we create both 32-bit and 64-bit installers. We also want to keep the download size small, so we do not generate a combined installer for both architectures.

Currently we built both 32-bit and 64-bit versions by [building twice on 32-bit and 64-bit windows](https://github.com/digidem/mapeo-desktop/blob/46adb9d8a788c125d99d84c7a1e38bd8977ac2c1/.github/workflows/nodejs.yml#L18-L21).

The way [Electron Builder](https://electron.build) [updater](https://github.com/electron-userland/electron-builder/tree/master/packages/electron-updater) works, it generates a file for the windows release called `latest.yml`. When we run the build on both 32-bit and 64-bit windows, they both generate this same file, and then upload it to Github releases and an S3 bucket. Unfortunately this means that the last build to run will overwrite the `latest.yml` for the previous build.

The auto-updater that is now part of Mapeo will use this file to discover, download and install the latest version. Because this `latest.yml` could point to either the 32-bit or the 64-bit installer, this could break for users on 32-bit Windows machines (because the updater will download and try to install the 64-bit version). It could also result in degraded performance for users on 64-bit machines if they update to the 32-bit version.

I'm not sure what the fix is. It seems like [PR#5718](https://github.com/electron-userland/electron-builder/pull/5718) might fix support for multiple archs without doing two builds, which _might_ fix this problem. Worst case: we try to ensure that the 64-bit version is always written to `latest.yml` and we turn-off auto-update for 32-bit Mapeo for now at runtime.

achou11:
Potentially fixes #492 

Leaving as a draft for now since there are still some things I need to figure out and test.

Notes:
- Upgrades `electron-builder` to `22.11.7` which has support for building multiple nsis arches in one step
  - requires an upgrade in the Node version to `14`. Haven't run into any issues with installation or running a simple dev environment with this version yet, but I also haven't thoroughly tested the various commands/workflows that would be affected by this
- unfortunately, since we build both the 32 and 64 bit arch, electron builder will also build the all-in-one installer as one of the artifacts. There doesn't seem to be a way to disable/bypass this. In the [docs](https://www.electron.build/configuration/nsis#32-bit-64-bit):

  > If you build both ia32 and x64 arch (--x64 --ia32), you in any case get one installer. Appropriate arch will be installed automatically. The same applied to web installer (nsis-web target).

- testing the publish step locally via `electron-builder --win --config ./builder.config.js --publish=never`, the generated `latest.yml` is a little unexpected in that it has duplicate keys for the 32 and 64 bit metadata (see output below). Not sure if this is something on our end, user-error, or a bug with the builder (i haven't found any relevant issues in the electron-builder repo). I don't think it's a major problem but definitely weird and ideally avoidable. Opened https://github.com/electron-userland/electron-builder/issues/6002 since I _think_ it's not something on our end.

```yml
version: 5.4.0-beta.12
files:
  - url: Installar_Mapeo_v5.4.0-beta.12_win.exe
    sha512: ...
    size: ...
  - url: Installar_Mapeo_v5.4.0-beta.12_win-x64.exe
    sha512: ...
    size: ...
  - url: Installar_Mapeo_v5.4.0-beta.12_win-x64.exe
    sha512: ...
    size: ...
  - url: Installar_Mapeo_v5.4.0-beta.12_win-x64.exe
    sha512: ...
    size: ...
  - url: Installar_Mapeo_v5.4.0-beta.12_win-ia32.exe
    sha512: ...
    size: ...
  - url: Installar_Mapeo_v5.4.0-beta.12_win-ia32.exe
    sha512: ...
    size: ...
  - url: Installar_Mapeo_v5.4.0-beta.12_win-ia32.exe
    sha512: ...
    size: 96826470
path: Installar_Mapeo_v5.4.0-beta.12_win.exe
sha512: ...
releaseDate: '2021-06-25T17:36:57.370Z'
```

- haven't been able to test if this actually fixes the autoupdating issue mentioned in #492 but the presence of several arch metadata in the generated `latest.yml` leads me to believe that it would be addressed by this change. I have VirtualBox set up for Windows but I need to make a proper testing strategy for autoupdate. I've done this in the past so I know the general steps I need to take but the execution will be a little tedious
- things that need to be done before this can be confidently merged:
  - ~make adjustments to CI. I don't have the best understanding of all of the steps that happen via the action so definitely need some guidance on the necessary changes there~ DONE
  - adjust the required checks for CI. since we no longer specify the `arch`, we just need to change them to the following:
    - `build (windows-latest, main)`
    - `build (macos-latest, main)`
    - `build (ubuntu-latest, main)`
  - ~ensure that important dev processes and workflows are not broken as a result of the Node version upgrade~ DONE
  - have some proof that the autoupdating issue is potentially fixed by this


builder.config.js
```diff
@@ -23,8 +23,8 @@ const config = {
     output: 'dist/main'
   },
   win: {
-    target: 'NSIS',
-    artifactName: 'Installar_Mapeo_v${version}_${os}-${env.ARCH}.${ext}',
+    target: [{ target: 'nsis', arch: ['x64', 'ia32'] }],
+    artifactName: 'Installar_Mapeo_v${version}_${os}-${arch}.${ext}',
```
achou11, line 27: since the new version of `electron-builder` actually handles multiple arches properly, we can use its built-in `arch` variable here instead

package.json
```diff
@@ -10,8 +10,9 @@
     "email": "info@ddem.us"
   },
   "main": "index.js",
-  "engine": {
-    "node": "12.14.1"
+  "engines": {
+    "node": "^14.17.1",
+    "npm": "^6.13.4"
```
achou11, line 15: wasn't sure if the `npm` version should be specified but figured it'd be helpful given that v7 is generally available

variants/icca/builder.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = async function (config) {
     artifactName: 'Mapeo_ICCAs_v${version}_${os}.${ext}',
     win: {
       ...config.win,
-      artifactName: 'Mapeo_ICCAs_v${version}_${os}-${env.ARCH}.${ext}'
+      artifactName: 'Mapeo_ICCAs_v${version}_${os}-${arch}.${ext}'
```
achou11, line 13: same comment as https://github.com/digidem/mapeo-desktop/pull/493/files#r659033856

package.json
gmaclennan: Good idea. We have not got around to moving to npm v7 yet. It's been quite annoying, because npm v7 updates all the lock files, which happened to me when my npm_merge_driver was using v7 while the project was using v6. Let's discuss package managers at some point, e.g. pnpm vs. npm vs yarn

package.json
```diff
@@ -10,8 +10,9 @@
     "email": "info@ddem.us"
   },
   "main": "index.js",
-  "engine": {
-    "node": "12.14.1"
+  "engines": {
+    "node": "^14.17.1",
```
gmaclennan, line 14: I've been pinning this to the same version of Node that the Electron version we are using runs, so that if we run tests for the backend code, then they are running in the same environment as they do in the app.

gmaclennan:
Can you remove the patch in `patches/`. This was in place because electron-builder did not recognize the tag when running in a Github Action. I _think_ that is fixed now, although I cannot find details in the electron-builder issues. Anyway, I think a better way to do this now is to include the publish / don't publish logic in the action config, rather than patching electron-builder code.

package.json
achou11: Hmm I see. Let me try downgrading back to 12 and seeing if the new version of builder still works. I only upgraded because of the upgraded dependency (there was a warning on installation for it but may not be a strict one)

achou11: Alright confirmed that upgrading node to 14 is necessary for the latest version of electron-builder (see https://github.com/electron-userland/electron-builder/issues/5953)

gmaclennan: ok, I'm fine with doing this engines upgrade. We don't currently have unit tests for the main process yet (the one that runs in Node 12.14.1 in electron), but if we do add them then we should ensure that we run the unit tests on Node 12.14.1, and keep the build process in Node 14.

These node versions are important because we have seen our stack break with patch versions of node, and we have not tested the Mapeo Core stack with Node 14 yet. We found that if we did not test with the exact version we're using at runtime, we got unexpected errors not caught by the tests.

At some point we need to update electron, which includes the Node@14 upgrade for the runtime, but we need to test Mapeo Core in Node 14 for that.

achou11: Yeah that make sense. Definitely not ideal to just upgrade the version for one dep but may be necessary because the most recent compatible version that potentially addresses our original issue is [`22.11.3`](https://github.com/electron-userland/electron-builder/releases/tag/v22.11.3) but using the `{arch}` macro that it enables is broken in that version but fixed in [`22.11.4`](https://github.com/electron-userland/electron-builder/releases/tag/v22.11.4), which we can't use unless we upgrade to node 14 🙃

.github/workflows/nodejs.yml
```diff
@@ -30,10 +29,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
-      - name: Use Node.js 12.14.1
+      - name: Use Node.js 14.X.X
         uses: actions/setup-node@v2
         with:
-          node-version: 12.14.1
+          node-version: 14
```
achou11, line 35: Should we pin the version here (current latest is `14.7.1`

gmaclennan: I guess if we want to ensure builds continue to work and are not affected by a bug in a Node release. Yes, let's pin.

gmaclennan:
So the one thing that I'm unclear on here is how/if the electron-updater works with these different Windows architectures? There does not seem to be enough information in the `latest.yml` for the updater to know which updater to use (because the only info is the filename, size and hash, and we control the filename through the electron-builder config, so it is not parsable, and I can't find any logic in the electron-updater code to check architecture). However I'm not sure at all how electron-builder uses this info in the `files` array to decide which updater to download? I see also your bug report about repeat entries.

I _think_ the best option might be a solution along the lines of https://github.com/electron-userland/electron-builder/issues/5087, however I think we can do this at build time rather than runtime by defining the `path` of the [S3Options](https://www.electron.build/configuration/publish#s3options) based on arch, then I _think_ builder will generate a `app-update.yml` (embedded in the packaged app) that points to that path, and hence use that path to look for `latest.yml` or `beta.yml` when doing auto-update.

I think the safest option for all of this might be to run each platform _and_ arch in their own CI build. There's not a huge cost for us: Github Action jobs can run in parallel and they are free. One concern is that even if building both ia32 and x64 together works right now, it might not work if we ever add a package that does not support prebuild (see https://www.electron.build/multi-platform-build.html). I think the original setup with the `npm_config_arch` option ensures that everything is built for the correct platform.

To test any of these theories we need a test publish setup, which is tricky because it involves creating tags in the repo. I suggest we set up a fork of `digidem/mapeo-desktop` e.g. `digidem/mapeo-desktop-test` (not sure it Github will fork to the same owner, but we can do that in git) and use that to test this publish setup. I can create an S3 bucket for publishing these test releases, then we can test the whole publish and auto-update workflow.

gmaclennan:
[this issue](https://github.com/electron-userland/electron-builder/issues/5922) has me concerned that there are bugs in this multi-arch code for windows in electron-updater, and from looking through the codebase there does not seem to be logic for properly handling this, which can break things.

achou11:
@gmaclennan all of your feedback is valid and after looking around their source code, I'd have to agree with you that we need to take a different path for this. Just to answer your questions:

> So the one thing that I'm unclear on here is how/if the electron-updater works with these different Windows architectures?

Yeah you're right that it isn't immediately clear if they handle this properly. I did find a couple of spots that hint that they at least attempt to do this properly. They have a type definition for Windows-specific update info [here](https://github.com/electron-userland/electron-builder/blob/71d37044d348ab0cd8f0f9f7754bf6ebaa4b996c/packages/builder-util-runtime/src/updateInfo.ts#L78-L86), which is then used in their [generic Provider handler](https://github.com/electron-userland/electron-builder/blob/28cb86bdcb6dd0b10e75a69ccd34ece6cca1d204/packages/electron-updater/src/providers/Provider.ts#L147-L156) when attempting to figure out which files from the provider to look at for the updated version. Again, this is just my guess so not fully confident in my interpretation

> I think the best option might be a solution along the lines of electron-userland/electron-builder#5087, however I think we can do this at build time rather than runtime by defining the path of the S3Options based on arch, then I think builder will generate a app-update.yml (embedded in the packaged app) that points to that path, and hence use that path to look for latest.yml or beta.yml when doing auto-update.

Yeah I like this workaround, especially if it can be achieved at build time. So we'll have to fix up the upload URL that we publish to for the Windows builds, based on architecture?

> I think the safest option for all of this might be to run each platform and arch in their own CI build. There's not a huge cost for us: Github Action jobs can run in parallel and they are free. One concern is that even if building both ia32 and x64 together works right now, it might not work if we ever add a package that does not support prebuild (see https://www.electron.build/multi-platform-build.html). I think the original setup with the npm_config_arch option ensures that everything is built for the correct platform.

Makes sense! Curious if we need to set more env variables based on this [page](https://github.com/electron/electron/blob/v9.4.4/docs/tutorial/using-native-node-modules.md#using-npm) from Electron's docs?

> To test any of these theories we need a test publish setup, which is tricky because it involves creating tags in the repo. I suggest we set up a fork of digidem/mapeo-desktop e.g. digidem/mapeo-desktop-test (not sure it Github will fork to the same owner, but we can do that in git) and use that to test this publish setup. I can create an S3 bucket for publishing these test releases, then we can test the whole publish and auto-update workflow.

That sounds like a proper way of doing this. From my past experience, since the updater only uses that to determine whether or not an update is available my dirty solution for testing this was to just adjust the version in `package.json` to something less than the currently published/available release and build it locally (or in CI) and install that build. So depending on what we want to test, that could be a low-code solution to testing if the autoupdater works or not

achou11:
Closing in favor of https://github.com/digidem/mapeo-desktop/pull/495

PR #495, opened 2021-06-30
Use separate autoupdate feed url for Windows 32-bit build

### LINKED ISSUE #492: Auto-updater does not consistently work with Windows 64-bit and 32-bit builds

achou11:
Fixes #492 

Notes:
- Discussed with @gmaclennan offline about the approach and decided on the following:
  - if building for Windows 32-bit, point the publish + autoupdate target url to the `/desktop-ia32` directory in S3 (or whatever we decide to call it)
  - pin `electron-builder` version to the most recent version that supports Node v12
- Updates the `electron-updater` version to the one that was released with the builder version we're using
- Updates the `electron-publish` patch to match the new version that's being used


.github/workflows/nodejs.yml
```diff
@@ -71,7 +69,7 @@ jobs:
           MAPEO_VARIANT: ${{ matrix.variant }}
       - name: Cleanup artifacts
         run: |
-          npx rimraf@2 'dist/${{ matrix.variant }}/!(*.exe|*.deb|*.AppImage|*.dmg)'
+          npx rimraf@2 'dist/${{ matrix.variant }}/!(*.exe|*.deb|*.AppImage|*.dmg|*.yml|github)'
```
achou11, line 72: We want to retain generated yaml files and `github/` directory for debugging purposes

builder.config.js
```diff
@@ -54,7 +54,7 @@ const config = {
     {
       provider: 's3',
       bucket: 'downloads.mapeo.app',
-      path: '/desktop'
+      path: process.env.ARCH === 'ia32' ? '/desktop-ia32' : '/desktop'
```
achou11, line 57: Did a local build with this and confirmed that the generated `app-update.yml` file had the following:

```yml
path: /desktop-ia32
```

package.json
```diff
@@ -110,7 +110,7 @@
     "electron-is-dev": "^1.1.0",
     "electron-rabbit": "^1.0.3",
     "electron-timber": "^0.5.1",
-    "electron-updater": "^4.2.2",
+    "electron-updater": "^4.3.9",
```
achou11, line 113: Wasn't sure if we should pin this but decided against doing so since I don't think it's strictly tied to the `electron-builder` version

builder.config.js
achou11: Also open to changing the directory name if there are any suggestions

.github/workflows/nodejs.yml
```diff
@@ -42,8 +42,6 @@ jobs:
           retry_on: error
           retry_wait_seconds: 0
           command: npm ci
-        env:
-          npm_config_arch: ${{matrix.config.arch}}
```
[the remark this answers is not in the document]
gmaclennan: I don't _think_ we need it, but no need to remove it if it's not affecting anything else.

builder.config.js
gmaclennan:
```suggestion
      path: process.env.ARCH === 'ia32' ? '/desktop-win-ia32' : '/desktop'
```

Since this is specifically windows (whereas the other path has all platforms)

gmaclennan (approved):
Looks good. thanks for this Andrew!

.github/workflows/nodejs.yml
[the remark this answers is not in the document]
achou11: i'll restore it in that case

PR #529, opened 2021-09-23
fix: Restore default tile server port to 5005

### LINKED ISSUE #527: Can't get Custom background maps to appear in Territory view
**Describe the bug**
Previously used this phrase to get custom maptiles to appear in Territory view, placed in box when you click on RHS button for background layers and then 'custom': http://localhost:5005/tiles/sinangoe-offline-raster-tiles-12/{zoom}/{x}/{y} with the tiles in a tiles folder within the Mapeo folder in Library. You would then refresh the screen and the background map should appear. This no longer works. I have tried with both png and jpg maptiles, and have tried closing and reopening Mapeo. 

Raised by @aliya-ryan  : https://www.notion.so/digidem/No-longer-possible-to-get-Custom-background-maps-to-appear-in-Territory-view-2a80b6023a8f4f2fb97445d3cd6cbeeb

QA Script - MAPEO Desktop v5.4.0-beta15 Territory : https://www.notion.so/digidem/QA-Script-MAPEO-Desktop-v5-4-0-beta15-Territory-Aliya-826de014a1c5401b8124c2d7283a455f#5e3dc098eda441b48dbfaa827c9d256b

Note:
Important to do before the next MD release

achou11:
Fixes #527

The default tile server port was changed to `5100` in 5467c19dddadd4d45391b8e9a7df7975d60d3919 :

https://github.com/digidem/mapeo-desktop/blob/5467c19dddadd4d45391b8e9a7df7975d60d3919/index.js#L27

Not entirely sure if that was intentional but for the sake of restoring previously working behavior and avoiding updates to programmatic documentation, this commit restores the default port to `5005`

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,7 @@ var argv = minimist(process.argv.slice(2), {
   default: {
     port: 5000,
     datadir: path.join(userDataPath, 'kappa.db'),
-    tileport: 5100,
+    tileport: 5005,
     mapPrinterPort: 5200
   },
   boolean: ['headless', 'debug'],


gmaclennan:
Oh, good find! I changed this because in https://github.com/digidem/mapeo-desktop/pull/465 I switched to using [`get-port`](https://www.npmjs.com/package/get-port), which gets the next available port if the requested port is in-use (as opposed to just crashing, which is what happens now). I switched it to `5100` to leave "room" for the Mapeo Server port, which defaults to `5000` (`get-port` keeps incrementing by `1` until it finds a free port).

As such, I think this change is fine and the best fix. There are two edge-cases:

- If the user's machine has something running on ports `5000`, `5001`, `5002`, `5003` and `5004`, then both the Mapeo Server and Tile Server will try to run on port `5005` (because we currently get free ports in parallel and do not lock them until we start the server listening). This is _very_ unlikely. The consequence will be a fatal error in the app (because the server will not be able to start)
- Not a consequence of this change, but a result of the #465 refactor: if port `5005` is in-use on the users machine, then the tile server will run on the next available port (e.g. 5006), so setting a custom background to use `5005` will not work.

I don't think these are significant to worry about at this stage, and when we refactor some of this (run a single server on a single port; run a map server with UI integration to avoid typing in template URLs) then this will be fixed in a more complete way.

gmaclennan (approved):

achou11:
@gmaclennan makes sense! I figured there was a strong reason to switch the ports. This wouldn't be as much of an issue if the documentation for programs was aligned with the change, but given the unlikelihood of the edge cases you mentioned, I'll just merge this and we can address later.

Also love how I did a whole git bisect (I think with ~20 steps) to find this one line 😅

PR #531, opened 2021-09-28
chore: Disable windows signing temporarily

achou11:
Waiting on new windows cert to unblock failing CI, so just need to build the assets in CI here so that programs can QA 5.4.0-beta.16


gmaclennan:
I'd lean towards merging this, so that all beta releases are on the main branch. We can highlight the lack of signing in the release notes, and turn it back on when we have the cert. I'm concerned that the new cert is taking time.

PR #534, opened 2021-09-29
chore: Add Husky to project

achou11:
Notes:
- [Husky](https://typicode.github.io/husky/#/) is a library that eases the configuration of git hooks for projects.
- Added a simple pre-commit hook that runs the `extract-messages` script. Note that this will not automatically add+commit any newly extracted messages. My thinking is that it's okay to be left with the choice to either add the generated changes as a subsequent commit or as an amendment to the previous commit.
- Other potential pre-commit tasks we'll want to add in the future include: linting/formatting staged files (with [`lint-staged`](https://github.com/okonet/lint-staged)), running tests


package.json
```diff
@@ -279,5 +283,11 @@
     "blocking": false,
     "logInfo": false,
     "logError": true
+  },
+  "lint-staged": {
+    "*.{js,ts,tsx}": [
+      "eslint --cache --fix",
+      "prettier-standard"
```
gmaclennan, line 290: I've used [`pretty-quick`](https://www.npmjs.com/package/pretty-quick) for this before, but just out of habit without actually researching if it is quicker. However since we are using prettier-standard here perhaps that won't work.

package.json
```diff
@@ -279,5 +283,11 @@
     "blocking": false,
     "logInfo": false,
     "logError": true
+  },
+  "lint-staged": {
+    "*.{js,ts,tsx}": [
+      "eslint --cache --fix",
```
gmaclennan, line 289: I'm not a use fan of linting in a commit-hook. It can be frustrating for developers when there are issues getting the tooling set up, or WIP that does not yet lint. Can we put this in a CI check instead?

package.json
achou11: I tried using pretty-quick without any further configuration and it seemed to completely ignore the rules specified by standard (even saving files in VSCode started format using prettier's default configurations)

package.json
achou11: yeah for sure! i'll leave it out of the the pre-commit config

achou11: just to clarify, are you just referring to the `eslint` step here or `lint-staged` as a whole?

ErikSin (approved):

PR #617, opened 2021-11-04
fix: Correct spelling for name of builds

### LINKED ISSUE #508: Correct spelling of 'Installer' in build output filenames
**Describe the bug**
We are currently naming build output files for all operating systems with the prefix "Installar_Mapeo_v$". Change prefix to "Installer_Mapeo_v$" in [builder config](https://github.com/digidem/mapeo-desktop/blob/master/builder.config.js) and update references in CONTRIBUTING.md (and index.js?)

achou11:
Fixes #508 

Notes:
- Can't think of any other things we need to update in order for this to safely go out, but definitely worth thinking through before releasing this


ErikSin (approved):
It looks good to me....maybe we should merge it and then try to trigger a release and see if anything blows up?

achou11:
Looks like there are a few potential places that need to be updated as a result of this:

- Minor: [mapeo-docs](https://github.com/digidem/mapeo-docs/search?q=installar), [mapeo-docs-new](https://github.com/digidem/mapeo-docs-new/search?q=installar)
- Important: [digital-democracy.org](https://github.com/digidem/digital-democracy.org/search?q=installar)

achou11:
Hm one thing I just thought of is that this could potentially mess with the auto-updater. I _think_ it's based on file-paths so the name change could cause older versions of desktop looking for `Installar` to not find anything to update to. Not entirely sure if that's how it works but that's my understanding. In that case, don't think there's much we can do besides suggesting them to install the release with the name change?

gmaclennan:
Can we change the name prefix to "Install_Mapeo" please? The spelling error was actually "Installar_Mapeo" --> "Instalar_Mapeo" since this was originally in Spanish and there was a typo, but I think now it makes sense to use an English filename for greater universal understanding now that Mapeo is used around the world.
Regarding the auto-updater, I believe it uses the filenames and URLs generated in `latest.yml` by electron-builder, which should use whatever filename we put in the config, so this should not affect that.

achou11:
> Can we change the name prefix to "Install_Mapeo" please?

Happy to change this 👍 

> Regarding the auto-updater, I believe it uses the filenames and URLs generated in `latest.yml` by electron-builder, which should use whatever filename we put in the config, so this should not affect that.

This PR does change the config though, so wouldn't that affect the generated filenames/URLs in `latest.yml`?

gmaclennan:
> This PR does change the config though, so wouldn't that affect the generated filenames/URLs in `latest.yml`?

Yes, I just checked the build files from the CD builds for this PR and the generated filenames in `latest.yml` are correctly updated to the new spelling. This is what we want: the previous version looks up the published `latest.yml` online, and uses that to generate the download URL for the update.

achou11:
> > This PR does change the config though, so wouldn't that affect the generated filenames/URLs in `latest.yml`?
> 
> Yes, I just checked the build files from the CD builds for this PR and the generated filenames in `latest.yml` are correctly updated to the new spelling. This is what we want: the previous version looks up the published `latest.yml` online, and uses that to generate the download URL for the update.

Ah got it. For some reason I thought the latest.yml was local to the installation but that makes sense 👍

bin/release_latest.js
```diff
@@ -32,10 +32,10 @@ gh.readFile(config.filename, { ref: config.branches[0] }, function (err, data) {
...
-  var windowsLine = `/mapeo/latest/windows https://github.com/digidem/mapeo-desktop/releases/download/v${version}/Installar_Mapeo_v${version}_win-x64.exe 302`
-  var win32Line = `/mapeo/latest/win32 https://github.com/digidem/mapeo-desktop/releases/download/v${version}/Installar_Mapeo_v${version}_win-ia32.exe 302`
-  var linuxLine = `/mapeo/latest/linux https://github.com/digidem/mapeo-desktop/releases/download/v${version}/Installar_Mapeo_v${version}_linux.deb 302`
+  var macLine = `/mapeo/latest/mac https://github.com/digidem/mapeo-desktop/releases/download/v${version}/Install_Mapeo_v${version}_mac.dmg 302`
+  var windowsLine = `/mapeo/latest/windows https://github.com/digidem/mapeo-desktop/releases/download/v${version}/Install_Mapeo_v${version}_win-x64.exe 302`
+  var win32Line = `/mapeo/latest/win32 https://github.com/digidem/mapeo-desktop/releases/download/v${version}/Install_Mapeo_v${version}_win-ia32.exe 302`
+  var linuxLine = `/mapeo/latest/linux https://github.com/digidem/mapeo-desktop/releases/download/v${version}/Install_Mapeo_v${version}_linux.deb 302`
```
achou11, line 38: @gmaclennan is it okay to remove this file as a whole? Just realized that it's not used anywhere, so not sure if it was part of a some local flow or something else

gmaclennan: Yes, it can be removed. Although this was the script that updated the website download links, which is why @ErikSin needed to manually update the website for the last release. I have now updated our website to use the permalinks for the latest releases which are provided by our [mapeo-releases-worker](https://github.com/digidem/mapeo-releases-worker)

PR #629, opened 2021-11-22
fix: Upgrade electron builder to latest version

### LINKED ISSUE #625: Can't start app from ZIP on MacOS
**Describe the bug**
Installing Mapeo via the `zip` file prevents it from being able to open on MacOS.

1. If you're on a Mac, you can go to the [downloads page](https://www.digital-democracy.org/mapeo/download/). Otherwise, install from https://github.com/digidem/mapeo-desktop/releases/download/v5.4.0/Installar_Mapeo_v5.4.0_mac.zip
2. Go through installation steps
3. Try to open application
4. See native error dialog

**Expected behavior**
You should be able to open the application after installation

**Screenshots**
image

**Desktop:**
 - OS: MacOS Big Sur
 - Version: 11.6.1
 
**Mapeo Version:**
 - Desktop 5.4.0

**Additional context:**
Looks like the version of `electron-builder` we're using has a relevant issue that was fixed in a later version. Hopefully, just upgrading versions will resolve this issue.
- original issue: https://github.com/electron-userland/electron-builder/issues/4299
- release with fix: https://github.com/electron-userland/electron-builder/releases/tag/v22.11.9

achou11:
Fixes #625 (see [builder changelog](https://github.com/electron-userland/electron-builder/releases/tag/v22.11.9))

Tested by successfully installing the associated MacOS `zip` asset generated in [CI](https://github.com/digidem/mapeo-desktop/actions/runs/1491755119)

Notes:
- **Important**: Upgrades the node version used in **development** from 12 to 14, which is required for this dependency upgrade. This **does not** affect the node runtime that's used within Electron itself, which still remains on 12 in this case

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -71,7 +71,7 @@ jobs:
           MAPEO_VARIANT: ${{ matrix.variant }}
       - name: Cleanup artifacts
         run: |
-          npx rimraf@2 'dist/${{ matrix.variant }}/!(*.exe|*.deb|*.AppImage|*.dmg|*.yml|github)'
+          npx rimraf@2 'dist/${{ matrix.variant }}/!(*.exe|*.deb|*.AppImage|*.dmg|*.yml|*.zip|github)'
         shell: bash
       - name: Upload artifacts
         uses: actions/upload-artifact@v1
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -30,10 +30,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
-      - name: Use Node.js 12.14.1
+      - name: Use Node.js 14.18.X
         uses: actions/setup-node@v2
         with:
-          node-version: 12.14.1
+          node-version: 14.18
       - name: npm ci
         uses: nick-invision/retry@v2
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/nodejs.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           MAPEO_VARIANT: ${{ matrix.variant }}
       - name: Cleanup artifacts
         run: |
-          npx rimraf@2 'dist/${{ matrix.variant }}/!(*.exe|*.deb|*.AppImage|*.dmg|*.yml|github)'
+          npx rimraf@2 'dist/${{ matrix.variant }}/!(*.exe|*.deb|*.AppImage|*.dmg|*.yml|*.zip|github)'
```
achou11, line 74: initially just changed this so I could actually download the asset from CI for testing this change, but maybe it's just worth keeping anyways

.github/workflows/nodejs.yml
```diff
@@ -30,10 +30,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
-      - name: Use Node.js 12.14.1
+      - name: Use Node.js 14.18.X
         uses: actions/setup-node@v2
         with:
-          node-version: 12.14.1
+          node-version: 14.18
```
achou11, line 36: Think it should be okay to not be specific about patch version but happy to change if necessary

Issue #452, opened 2020-10-20
menù shifting


anthos-github:
pressing the point menù, the area drawing were acrivated, not the point .

![image](https://user-images.githubusercontent.com/73160598/96554136-6d2dd480-12b6-11eb-8a66-54dd183f85ae.png)

rudokemper:
Just providing some more context here: this is what you see when you run the Mapeo walkthrough (as prompted upon first load of the app, or you can run it again by going to `Help` and then `Start the Walkthrough` in Territory view). This walkthrough as a whole is broken with things appearing out of place, and has not been customized to Mapeo - it's a direct fork of the iD walkthrough which is about OSM editing. 

I would propose hiding the `Help` menu, and deactivating the prompt for starting the walkthrough upon first load, until these are fixed and customized for Mapeo specifically.

PR #635, opened 2021-12-03
fix: hide unhelpful iD Help pane until properly customized for Mapeo

rudokemper:
### Contributor checklist:

- [X] My contribution is **not** related to translations. _Please submit translation changes via our [Mapeo Crowdin project](https://crowdin.com/project/mapeo-desktop)_
- [X] My commits are in nice logical chunks with [good commit messages](http://chris.beams.io/posts/git-commit/)
- [X] I have walked through the [QA Manual Testing
- [ ] If my changes depend upon an update to a dependency, I have updated the package-lock.json file using `npm install --package-lock`

### Description

Partial fix for #452 although there are other things to address. 

This PR hides the `Help` button in Mapeo Desktop Territory, since the content of the Help pane that it opens is leftover from iD Editor / OSM and has not at all been customized for Mapeo. It is more confusing than helpful. It also provides a link back to the Walkthrough which is buggy and also not customized for Mapeo.

Trainers from the field report back that they have to tell their users to ignore this pane, which does not inspire confidence. So until we have customized the content, let's hide it for now.

There are still two remaining tasks to fully address disabling this inherited feature, however:

1) The `H` key still opens this pane. This needs to be deactivated.
2) Upon first load of Mapeo, the user is prompted to do a Walkthrough from iD / OSM. I recommend disabling this as well until the buggy aspects of it (i.e. offset transparency overlays) it's been customized for Mpaoe.

image

### DIFF
--- a/src/renderer/components/MapEditor/index.js
+++ b/src/renderer/components/MapEditor/index.js
@@ -112,7 +112,10 @@ insertCss(`
   }
   .id-container .layer-list li.switch {
     background: none;
-  }  
+  }
+  .id-container .map-controls .help-control {
+    display: none;
+  }
   .id-container .map-panes .map-data-photo-overlays {
     display: none;
   }


rudokemper:
👀 @ximenabb

achou11 (approved):
The two other issues are things that would have to be done to the id editor fork we have right? not a blocker for this PR imo though

rudokemper:
@achou11 yeah (1) for sure is, maybe worth indexing as to-do for that repo.
(2) I haven't scoped out yet so not sure where the relevant code resides.

achou11:
> @achou11 yeah (1) for sure is, maybe worth indexing as to-do for that repo. (2) I haven't scoped out yet so not sure where the relevant code resides.

makes sense. from a brief look it seems that (2) is also built into iD. And yeah we should probably create issues on https://github.com/digidem/iD-mapeo that are based on the PRs you've been creating here, for the sake of at least tracking and potentially implementing if straightforward enough

rudokemper:
Makes sense -- will do!

PR #637, opened 2021-12-08
chore: Made closing text translatable

ErikSin:
# Closing Text Translatable
In Message/Main/, closing string added. This string is added, via the DOM, to the closing.html.

When language is changed , ipcRenderer('set-locale') is fired, this sets off a fs.writeFile(), in the main process, which writes the closing string to a JSON file. Then the JSON file is read in a preload script for closing.html, and is appended to the DOM.

Kapture 2021-12-10 at 13 07 11

### DIFF
--- a/messages/main/en.json
+++ b/messages/main/en.json
@@ -101,5 +101,6 @@
   "convert-button": "Copy to Editor",
   "convert-number": "You have filtered your data to show {count} observations",
   "convert-detail": "{count} of these are not yet in Map Editor. Click OK to move these {count} observations to Map Editor.",
-  "convert-nothing": "These are all already in the Map Editor."
+  "convert-nothing": "These are all already in the Map Editor.",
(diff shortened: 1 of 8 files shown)


achou11:
Did an implementation of this using just IPC so as to avoid creating any intermediary files. LMK what you think of it

https://github.com/digidem/mapeo-desktop/commit/415a1a3de4fd06acc4cfcfe6d148431ac7632a67

ErikSin:
> Did an implementation of this using just IPC so as to avoid creating any intermediary files. LMK what you think of it
> 
> [415a1a3](https://github.com/digidem/mapeo-desktop/commit/415a1a3de4fd06acc4cfcfe6d148431ac7632a67)
> 
> can put up as separate PR if desired

I think it would probably makes the most sense to make this into a PR

achou11:
Closing in favor of #640

PR #678, opened 2022-04-25
Background Map Management

ErikSin:
## Background Map Management

Screen built for managing all background maps/styles

Screens to view all background map styles, and see all the offline areas associated with each style.

Button to delete style (button with no functionality)

https://user-images.githubusercontent.com/67773827/165188789-29f6dd97-8630-4a97-83be-d778ffcf4778.mp4

## Figma:
https://www.figma.com/file/sxz2NvysPVab6ZTqs6f57K/Cobra-Collective?node-id=245%3A1327

## To Do:
- Api Calls


src/renderer/components/BackgroundMaps/MapCard.js
```diff
@@ -0,0 +1,104 @@
...
+        style={{ backgroundColor: !isBeingViewed ? '#CCCCD6' : '#0066FF' }}
+      >
+        <div style={{ width: '30%' }}>
+          <Mapbox
+            containerStyle={{
+              height: '100%',
+              width: '100%'
+            }}
+            style={offlineMap.url}
+          />
```
ErikSin, line 60: Does this look correct? Currently it is not showing any style, just a green box

achou11: this seems correct! curious about what dataset you're using for your example.

perhaps unrelated, but will note that the way you're instantiating mapbox-react-gl is a little odd. Usually you'd create the instance outside of react i.e. not in a component like you're doing in line 30. See the example from the docs: https://github.com/alex3165/react-mapbox-gl#getting-started

gmaclennan: Yes creating an instance of Mapbox-gl is expensive (which is what creating an instance of mapbox-react-gl will do) so you should use a single instance for all maps rather than a different instance for each map.

ErikSin: So I created one instance for my two components [here](https://github.com/digidem/mapeo-desktop/pull/678/commits/b9623128433155c6dcb90d5f78548a7f711b0a24). Should I extend that further and use that same instance for the `mapview.js` (which is actually instantiated in a  React compnonent and memoized as well, but fixing that is out of scope for this PR)

achou11: i wouldn't touch that one because

1) out of scope for this PR

2) Wondering if there's some context that i'm missing for that case to instantiate inline. Looks like it relies on a couple of props (basically the access token and the type of screen it's rendered in e.g. map printer), so i'm thinking that that was a pretty intentional (but not ideal) choice. seems like that could be refactored to be improved, but not worth doing in this PR since it's out of scope

src/renderer/components/BackgroundMaps/BackgroundMapInfo.js
```diff
@@ -0,0 +1,163 @@
...
+            <div style={{ padding: 20 }}>
+              <Typography variant='subtitle2' style={{ fontSize: 18 }}>
+                {data.name}
+              </Typography>
+              {data.zoom && (
+                <Typography variant='body1'>
+                  {t(m.zoomLevel, { zoom: data.zoom })}
+                </Typography>
+              )}
+              <Typography variant='body1'>{`${size} ${t(m.mb)}`}</Typography>
```
achou11, line 85: `size` needs to be converted to MB since it's in bytes up to this point

image

src/renderer/components/Settings/index.js
```diff
@@ -0,0 +1,139 @@
...
+ * @prop {function(boolean):void} setReset
+ * @prop {boolean} fadeIn
+ */
+
+/** @param {SettingsProp} props */
+export const Settings = ({ reset, setReset, fadeIn }) => {
+  const [menuVisible, setMenuVisibility] = React.useState(true)
+
+  /** @type {import('./SettingsMenu').SettingsTabs['tabId'] | false} */
+  const initialState = /** {const} */ (false)
```
achou11, line 49: would make this variable a little more descriptive or just inline it to the useState initializer.

src/renderer/components/Settings/index.js
```diff
@@ -0,0 +1,139 @@
...
+
+  const [tabValue, setTabValue] = React.useState(initialState)
+
+  const classes = useStyles()
+
+  if (reset) {
+    setReset(false)
+    if (tabValue === 'BackgroundMap') setTabValue(false)
+  }
```
achou11, line 58: this state resetting code isn't necessary because you remount this component whenever it's navigated to i.e. in `Home.js` you do `tabIndex === 4 && <Settings ... />`

you don't need the `const [settingsReset, setSettingsReset] = React.useState(false)` in `Home.js` as a result and can remove the associated props from this component in that case.

tangentially, wonder if it makes sense to not conditionally render the Settings tab (like we do for the other app tabs) because that way you preserve the state when returning to it, which may be desirable from a UX perspective (as opposed to always starting in the reset state when navigating to it). This behavior could be useful when importing a map and you want to navigate elsewhere in the app without losing anything about the map import you started.

src/renderer/components/Settings/SettingsMenu.js
```diff
@@ -0,0 +1,37 @@
...
+      value={currentTab}
+      onChange={(e, newValue) => setCurrentTab(newValue)}
+    >
+      {tabs.map(tab => (
+        <Tab
+          style={{ textTransform: 'none' }}
+          label={t(tab.label)}
+          key={tab.tabId}
+          value={tab.tabId}
+          icon={tab.icon}
```
achou11, line 32: Icon and label should be vertically aligned with each other:

image

achou11 (changes requested):
Are the fade animations part of the design spec? if not, would say to omit them from this PR because they feel unnecessary and we don't use them elsewhere in the app at the moment

src/renderer/components/BackgroundMaps/BackgroundMapInfo.js
```diff
@@ -0,0 +1,163 @@
...
+  id,
+  idBeingViewed,
+  unsetMapValue,
+  url,
+  size
+}) => {
+  const shouldLoad = React.useMemo(() => id === idBeingViewed, [
+    id,
+    idBeingViewed
+  ])
```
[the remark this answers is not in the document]
ErikSin: yes, its for the fade effect. If we conditionally render using the parent, we still need to pass it some boolean to indicated that the fade effect should be run. So since we are already passing in the boolean, it seems to just handle all the loading state inside the component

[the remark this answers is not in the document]
achou11: i see. aside from whether the fade effect is necessary at all, you don't need an explicit variable in this case to trigger the fade. specifying `<Fade in ... />` will work the same in this case.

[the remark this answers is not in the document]
achou11: the `BackgroundMaps` component could be updated as such as a result:

```tsx
export const BackgroundMaps = ({ openSettings }) => {
  const { formatMessage: t } = useIntl()

  /** @type {MapServerStyleInfo['id']|false} */
  const initialMapId = /** {const} */ (false)

  const [mapValue, setMapValue] = React.useState(initialMapId)

  const { data } = useMapServerQuery('/styles')

  function unsetMapValue () {
    setMapValue(false)
  }

  const offlineMap = data && data.find(m => m.id === mapValue)

  return (
    <React.Fragment>
      <SidePanel
        mapValue={mapValue}
        openSettings={openSettings}
        setMapValue={setMapValue}
      />

      {!mapValue || !data ? (
        <div style={{ padding: 40 }}>
          <Typography variant='h4'> {t(m.mapBackgroundTitle)}</Typography>

          <Typography variant='body1'>
            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
            eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim
            ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
            aliquip ex ea commodo consequat. Duis aute irure dolor in
            reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
            pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
            culpa qui officia deserunt mollit anim id est laborum.
            <br />
          </Typography>
        </div>
      ) : offlineMap ? (
        <BackgroundMapInfo
          key={offlineMap.id}
          size={offlineMap.bytesStored}
          id={offlineMap.id}
          unsetMapValue={unsetMapValue}
          url={offlineMap.url}
        />
      ) : null}
    </React.Fragment>
  )
}
```

src/renderer/components/BackgroundMaps/BackgroundMapInfo.js
ErikSin: https://github.com/digidem/mapeo-desktop/pull/678/commits/35536ba7d93b16a61bd742dc226353e375e9499c

src/renderer/components/BackgroundMaps/BackgroundMapInfo.js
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/mapeo-desktop/pull/678/commits/35536ba7d93b16a61bd742dc226353e375e9499c

src/renderer/components/Settings/index.js
ErikSin: So the reason for `setReset()` was if the user wanted to click "settings" again to reset. When background maps tab is open, it takes over the settings page completely. The `reset` variable made it possible for the user to click the settings tab again to go exit the background maps tab. That being said, if we don't want to conditionally render it, then we can't have both, since clicking "settings" will bring you back to where you were in the settings instead of resetting the state

src/renderer/components/Settings/index.js
ErikSin: This is actually a very annoying this about jsdoc. The typescript equivalent would be `useState<SettingsTabs['tabId']|false>` In order to do this with JSdoc, i need to initialize the first state of the `useState()` above it as I did here.

ErikSin: I renamed it so it is a little clearer: https://github.com/digidem/mapeo-desktop/pull/678/commits/73472a2872649cdd5aeff1f3772395226c25a63d

src/renderer/components/Settings/index.js
ErikSin: took out set reset and made the rendering NOT conditional: https://github.com/digidem/mapeo-desktop/pull/678/commits/73472a2872649cdd5aeff1f3772395226c25a63d

achou11: ah i see. makes more sense to me now

> if the user wanted to click "settings" again to reset.

is this necessary given that the background maps settings screen has a back button in the sidebar? just thinking that adding the additional logic to cover this use case may not be worth it for now if a mechanism already exists

ErikSin: it was slightly confusing to hit settings and see the background maps tab. But if we don't want to conditionally render it, then we have to get rid of `setReset()` (Or add even more complicated logic) anyways

achou11: just remembered that you can reset the react state by providing a new `key` prop value, so it may be possible to avoid adding extra code and just providing a new value for the Settings tab panel `key` prop in `Home`.

achou11: think it makes sense to avoid any reset behavior for now since it's less code to work with. if the reset stuff is desired, should be easy to add later on

ErikSin: just to clarify, we can't really have both the "saving of the tab state", and "resetting settings on click". Either we click `settings` and it goes to the saved tab state Or we click `settings` and it resets the tab state

achou11: yeah that's how I understood it and I'm leaning towards the former (preserve state) unless you're okay with adding the code to account for the latter

src/renderer/components/Settings/SettingsMenu.js
ErikSin: https://github.com/digidem/mapeo-desktop/pull/678/commits/c607d027aab4e6b44a0b17e6e74cb57fba736a3e

ErikSin: for what its worth, this was annoyingly hard to do with MUI ..... :joy:

PR #680, opened 2022-04-29
fix(obs): only show filters for fields in point presets

gmaclennan:
**As yet untested**. Pending testing from build.

A quick fix for an issue where the filter pane in observation view shows filters for fields that do not relate to data that would appear in observation view. E.g. filters will currently show for fields which are only used by presets for areas, but the observation view only displays point data. Also, currently filters will show for fields that are defined in a config file, but not used by any presets.

Caveat: In the future it might be better to filter which fields appear based on the dataset rather than the presets.

### DIFF
--- a/src/renderer/components/MapFilter/MapFilter.js
+++ b/src/renderer/components/MapFilter/MapFilter.js
@@ -135,9 +135,19 @@ function usePresets () {
       try {
         const data = await api.getPresets()
         const presets = mapToArray(data.presets)
-        const fields = mapToArray(data.fields)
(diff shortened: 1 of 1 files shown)


gmaclennan:
I tested the build locally and it appears to work, but I would appreciate some eyes on this to check I'm not making any silly mistakes.

gmaclennan:
**Before:**
Screenshot 2022-05-09 at 14 55 36



**After:**
Screenshot 2022-05-09 at 14 41 34

achou11 (approved):

Issue #652, opened 2022-01-20
(Front) Leave Practice Mode


ximenabb:
**Description**
User accepts a project invite, therefore leaving practice mode. 
If the user had created observations (min 1) while on practice mode, app asks if users wants to keep obvs.

Screen is:
Leave Practice Mode
Bring (XXX) observations from Practice Mode to the project?
No, delete my observations
Yes, keep my observations
(Cancel) (Leave Practice Mode)

**Copydeck:**
https://docs.google.com/spreadsheets/d/1JvC_zVxIJaZIsDyWfwZ7wl7jcSFNdgYBh-DWIVJr_eA/edit#gid=1667712736

**Design** 
https://www.figma.com/file/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=47881%3A1241

Issue #650, opened 2022-01-20
(Front) Project Invite / Receiving a project invite 


ximenabb:
Description :
- As with Mobile, Project Joiner is anywhere in MD and receives the Join Project invite
- MD can only receive a project invite or request  a project at the moment (not send an invite to join a project)
- MD and MM must be on the same network
- User can only receive an open if MD is open on their device

Screen is:
Join Project
Show this QR code to a project coordinator
(Send Join Request Instead) (Cancel)

**Design**
https://www.figma.com/file/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=47885%3A1423

PR #695, opened 2022-08-02
feat: Project Invite Flow (Screen Only) and Practice Mode

ErikSin:
## Description
2 new (but related) features:
- Practice Mode
- Invite to new Project flow (screens only)

### Practice Mode
When user is using the default config, they will be in practice mode. Practice Mode is indicated by an orange strip across the bottoms of the screen. 

### Invite to a new Project
With the upcoming refactor of Mapeo core users will have the have a more streamlined approach to join a project. 
Users will be able to request to join a project. In Mapeo Desktop, user will be able to join a project through a url or QR code (scanned by the project admin through Mapeo mobile).

## Screen Shots

**Settings Menu**
image

**Practice Mode**
image

**Join Project Flow**
image

image

image

## Linked Sprint Item
Closes #694 #650 #652 

## To Do
- [ ] Put Behind Feature Flag
- [ ] Approve Copy Deck
- [ ] Import Config button functionality


messages/renderer/en.json
```diff
@@ -432,6 +444,37 @@
...
+    "description": "Label to indicate the config",
+    "message": "Config"
+  },
+  "renderer.components.Settings.ProjectConfig.importConfig": {
+    "description": "Button to import a new config",
+    "message": "Import Config"
+  },
+  "renderer.components.Settings.ProjectConfig.joinProject": {
+    "description": "Button to join a new project",
+    "message": "Join A Project"
```
achou11, line 461: confirm that `A` is capitalized in the copy deck?

src/main/windows.js
```diff
@@ -87,6 +87,7 @@ function ClosingWindow () {
     frame: false,
     show: false,
     alwaysOnTop: true,
+    transparent: true,
```
achou11, line 90: what's the purpose of adding this?

src/renderer/components/Home.js
```diff
@@ -154,12 +166,7 @@ const StyledPanel = styled.div`
   }
 `
 
-const Version = styled.div`
-  align-self: flex-start;
-  margin: auto 10px 10px 10px;
-  font-size: 0.8rem;
-  color: ${buildConfig.variant === 'icca' ? '#eeeeee' : '#aaaaaa'};
-`
+// {buildConfig.variant === 'icca' ? '#eeeeee' : '#aaaaaa'};
```
achou11, line 169: assuming used for debugging?

src/renderer/components/Home.js
```diff
@@ -227,6 +234,31 @@ export default function Home ({ onSelectLanguage }) {
...
+
+  React.useEffect(() => {
+    api.getMetadata().then(metadata => {
+      if (metadata.name === 'mapeo-default-settings') {
+        setPracticeModeOn(true)
+        return
+      }
+      setPracticeModeOn(false)
+    })
+  }, [api, setPracticeModeOn])
```
achou11, line 260: no need for api to be in the deps array here, as the PR warning suggests

src/renderer/components/Home.js
```diff
@@ -227,6 +234,31 @@ export default function Home ({ onSelectLanguage }) {
   const [update, setUpdate] = useUpdater()
   const { formatMessage: t } = useIntl()
 
+  const [settingsReset, setSettingsReset] = React.useState(false)
+  const [practiceModeOn, setPracticeModeOn] = React.useState(false)
+  const [invite, setInvite] = React.useState(null)
+
+  const resetInvite = useCallback(() => {
+    setInvite(null)
+  }, [setInvite])
```
achou11, line 243: probably somewhat premature to useCallback here?

src/renderer/components/Settings/index.js
```diff
@@ -0,0 +1,127 @@
...
+    icon: <AssignmentIcon />,
+    label: m.projConfig,
+    subHeader: m.projConfiSubHeader
+  }
+]
+
+/**
+ * @typedef SettingsProp
+ * @prop {boolean} reset
+ * @prop {function(boolean):void} setReset
```
achou11, line 42: unused prop?

src/renderer/components/Settings/index.js
```diff
@@ -0,0 +1,127 @@
...
+  /** @type {SettingsTabs['tabId'] | false} */
+  const initialState = /** {const} */ (false)
+
+  const [tabValue, setTabValue] = React.useState(initialState)
+
+  const classes = useStyles()
+
+  return (
+    <SettingsProvider practiceModeOn={practiceModeOn}>
+      <Fade in={reset || fadeIn} timeout={FADE_DURATION}>
```
achou11, line 61: seems like the only usage of `reset` and `fadeIn` props - perhaps we can merge as a single prop to make it less work for the parent components?

src/renderer/components/dialogs/ProjectInvite.js
```diff
@@ -0,0 +1,192 @@
...
+
+  const [migrationPlanConfirmed, setMigrationPlanConfirmed] = React.useState(
+    false
+  )
+
+  const classes = useStyles()
+
+  if (practiceModeOn && !migrationPlanConfirmed) {
+    return (
+      <React.Fragment>
```
achou11, line 81: probably has minimal effect but just be somewhat weary of using fragments to wrap entire components because it subjects it to being less predictably styled by the parent. not sure if an issue in this case though

src/renderer/components/Home.js
```diff
@@ -256,50 +288,85 @@ export default function Home ({ onSelectLanguage }) {
...
+          <StyledTabs
+            style={{ height: '100%' }}
+            className={classes.root}
+            orientation='vertical'
+            variant='scrollable'
+            value={tabIndex}
+            onChange={(e, value) => {
+              if (value === 4) setSettingsReset(true)
+              setTabIndex(value)
+            }}
```
achou11, line 310: a little bit hard to visualize what's going on here. when does `value === 4`? is there a more descriptive/explicit way to fulfill the same condition?

achou11 (commented):
Overall looks solid so far! left some relatively minor comments but can re-review when remaining items are addressed

src/renderer/components/Home.js
ErikSin: I suppose it is premature, but we should eventually use `usecallback` here when we deal with the invite. I don't think adding it now adds to much overhead, and I think it will be a good reminder for the future.

src/renderer/components/Home.js
ErikSin: Sort of. I can explicitly name the tab. And the i can say if ``value` === that name. But there is no type safety to it, and all the other tabs are using the index. I want to change this component to have type safety....but styled components is [not letting that happen](https://stackoverflow.com/questions/71855570/styled-components-could-not-be-used-as-jsx-element-react-typescript):

Ill change it to be more explicit, but this whole page (or all of desktop realistically )....

src/renderer/components/dialogs/ProjectInvite.js
ErikSin: These component are specifically going to be wrapped in a dialog (they are not exported components, so used very specifically), so i think in this case it is not an issue!

src/renderer/components/Home.js
achou11: ah i see, `4` is passed as an explicit value to one of the tabs. in that case, using a constant defined outside of the component would make this easier to understand at the very least. I thought the value was referring to the index position of the tab, which would've made it trickier and less reliable

src/renderer/components/Home.js
ErikSin: Actually im going to name this handleInvite...to be a little more explicit

src/renderer/components/Settings/index.js
ErikSin: Agreed, this will make sense later once we deal with the map component (the map component hides the tab bar, so having a reset available does make it easier). But ill get rid of it until we merge that in

src/main/windows.js
ErikSin: whoops, not needed

PR #708, opened 2022-11-28
Es/map server

ErikSin:
# Map Server Integration


src/renderer/hooks/useMapServerQuery.js
```diff
@@ -0,0 +1,22 @@
+// @ts-check
```
ErikSin, line 1: @gmaclennan if we didn't create an API layer, we would have to type each query as shown here

src/renderer/hooks/useMapServerQuery.js
```diff
@@ -0,0 +1,22 @@
...
+  return useQuery({
+    queryKey: [resourcePath],
+    queryFn: () => ky.get(MAP_SERVER_URL + resourcePath).json()
+  })
+}
+
+/**
+ * @type {import('@tanstack/react-query').UseQueryResult<ReturnType<import('@mapeo/map-server/dist/api/styles').StylesApi["listStyles"]>>}
+ */
+const { data } = useMapServerQuery('/styles')
```
ErikSin, line 22: if we don't create an API layer, we will need to type every `useMapServerQuery` @gmaclennan

src/renderer/hooks/useMapServerQueryWithImplicitTypings.js
```diff
@@ -0,0 +1,26 @@
...
+    const listSyles = ky.get(MAP_SERVER_URL + '/styles').json()
+    return listSyles
+  }
+
+  return {
+    getStyle: getStylesFn
+  }
+}
+
+const { data } = useQuery({
```
ErikSin, line 23: @gmaclennan with an API layer, the `data` type is implicitly defined by out type assertions in API layer

ErikSin:
@gmaclennan I created 2 hooks (technically 1 hook and an API layer) to show the difference in typescript typings. If we do create an API layer, we can control the typing a little easier. And react-query will infer the typings from the api type assertions.

If we don't have an API layer, we will need to assert the type to the hook everytime we use it, which gets a little clunky with `tanstack` typings. 

I know you didn't want to create that extra layer, so I am curious if you think there is a more efficient way of doing this?

src/renderer/hooks/useMapServerQuery.js
gmaclennan: I think it's best not to add more code just to solve a typing issue.

I think the easiest way to do this is with function overloads ([example](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAawJ5wGZQiOByZ2BQokscA3nAK4DOApgIoU1QoC+aGW2AAjAIYB2VPgGMEAeig1ewmAFoAjo2YF8+VBX4zgEfpVoBZXmADKTAG5MGTJAApJVCBSjCaABV4wAFgC4cYoUgANjRU2ACUvgCCUFC8SAA85MAAJr5CUMD8AOZwLAB8ahpaOno0hibmlkq29o7Obh4+cAAG-jBBIWIAJKTpmVkszRFkcPy8IDRpMBnZuYWaMNq61GVGplAWUFbMdiF1Lu5eUzNZwwIopPhwcJIwTsu027aX19eK1gDSNEi+ANq1TgOjQAugAaK6vd7MABi-F8NjCcAAvHlEEgAHRZGgwGzYTwwGBgbxiMSBCDCXiBTwQITeAAcAAZGdg4ABqG57QENLxhdEAKwc-AREJYYXwLFUwh0QjgAWCVAAMsAZUjSuV1psnri2h1QmKpYJ4HKaABJfioCDItVrSpbara41UMS8ABGwnC+ANMsyZkpKStK3Vtq12DEPr9yUOnnCcBJcHaYBCwgyYHgTAwUEl0ogwXRZKyNkdSqEvwZwPRKURcf4lvT0FB8aQibgytl036nuzufzhfawTNFvRYwmVbEo0tCaTKbTMXrjebrb62XwQA)).

I don't think this can be done in JSDoc though, so would require a typescript file to define the type and then import here.

src/renderer/hooks/useMapServerQueryWithImplicitTypings.js
gmaclennan: See above for a solution that doesn't require this file.

gmaclennan:
@ErikSin see my commit https://github.com/digidem/mapeo-desktop/pull/708/commits/70fad5ffda7014d34009551fd1cc1592d87defb3 for an example of how you might do this.

ErikSin:
> @ErikSin see my commit [70fad5f](https://github.com/digidem/mapeo-desktop/commit/70fad5ffda7014d34009551fd1cc1592d87defb3) for an example of how you might do this.

Unfortunately not working as expected. With `styles/${string}` I should be expecting a `styleJson`. But I am still getting a `MapServerStyleInfo`

image

Also getting a weird error, related to typing a function:
image

image

src/renderer/hooks/useMapServerQuery.js
ErikSin: I understand how to do this typescript, but I am not sure I fully understand how we would implement this is JSDOC. I would create a `ts` file define everything there, import it into a `js` file, then export it, almost like an alias? Is that what you are suggesting?

gmaclennan:
Indeed, it seems like function overloads are not really supported in JSDoc, which is a shame. I think the best way around this is a manual `d.ts` file for the hook. I've pushed an example to a separate branch: https://github.com/digidem/mapeo-desktop/commit/53ac077200e774cfcb34e2d1b56cb8796b453e3a

Verified locally by running:

```
❯ ./node_modules/.bin/tsc --noEmit --allowJs --esModuleInterop --jsx react src/renderer/hooks/MapServerExample.js
```

package.json
```diff
@@ -284,7 +288,7 @@
     "logError": true
   },
   "lint-staged": {
-    "*.{js,ts,tsx}": [
+    "*.{js,tsx}": [
```
ErikSin, line 291: prettier is not playing well with `.d.ts` files. I will look further into this, but I wanted to push this for review while I did that

src/main/index.js
```diff
@@ -5,6 +5,7 @@ const { app, dialog, MessageChannelMain, ipcMain } = require('electron')
 const isDev = require('electron-is-dev')
 const contextMenu = require('electron-context-menu')
 const mkdirp = require('mkdirp')
+const createMapServer = require('@mapeo/map-server')
```
achou11, line 8: since we're doing TS checking in this file, would update this import to:

```suggestion
const createMapServer = require('@mapeo/map-server').default
```

That way you get the proper TS annotation (see https://github.com/digidem/mapeo-map-server#usage)

src/main/index.js
```diff
@@ -96,6 +106,20 @@ async function startup ({
...
+  // Running this in the main thread rather than in a background process because
+  // map-server uses Node Worker Threads, and the background processes actually
+  // run in a browser window (with Node integration turned on) but they do not
+  // support Node workers (only Web Workers). The reason for running the other
+  // processes (map printer and mapeo core) in a background process was because
+  // processes in the main thread are blocking for the render thread.
+  // Fortunately map server does not have any expensive functions so it should
+  // not slow down the main process nor block the render thread if we run it
+  // here from the main process...
+  // @ts-ignore
```
achou11, line 118: can remove this if you update the import as suggested in the comment above

src/renderer/hooks/useMapServerQuery.js
```diff
@@ -0,0 +1,15 @@
...
+import ky from 'ky/umd'
+import { useQuery } from '@tanstack/react-query'
+
+// local host and global port number
+// @ts-ignore
+const MAP_SERVER_URL = 'http://127.0.0.1:' + window.mapServerPort
+
+export function useMapServerQuery (resourcePath, enabled) {
+  return useQuery({
+    queryKey: [resourcePath],
```
achou11, line 11: not super familiar with react query but based on the [docs](https://react-query-v3.tanstack.com/guides/query-keys) related to query keys, should this be some sort of partitioned array of items based on the path segments of the `resourcePath`? For example if we're trying to get a specific style `/styles/style-id-1`, we'd have something like `queryKey: ['styles', 'style-id-1']`.

Maybe doesn't matter for how we're using it right now, but just curious if that's the more idiomatic approach to the query key

src/renderer/hooks/useMapServerQuery.js
```diff
@@ -0,0 +1,15 @@
+// @ts-check
+import ky from 'ky/umd'
+import { useQuery } from '@tanstack/react-query'
+
+// local host and global port number
+// @ts-ignore
+const MAP_SERVER_URL = 'http://127.0.0.1:' + window.mapServerPort
```
achou11, line 7: more of a nit but guessing this is because `window.mapServerPort` is type `any`? maybe use `as` just so that it's clear what type the variable is supposed to be i.e. 

```suggestion
const MAP_SERVER_URL = 'http://127.0.0.1:' + (window.mapServerPort as number)
```

alternatively, could add a runtime check to make sure it's defined as a number and throw an exception if it's not

src/renderer/hooks/useMapServerMutation.js
```diff
@@ -0,0 +1,26 @@
+// @ts-check
+import ky from 'ky/umd'
+import { useMutation, useQueryClient } from '@tanstack/react-query'
+
+// @ts-ignore
+const MAP_SERVER_URL = 'http://127.0.0.1:' + window.mapServerPort
```
achou11, line 6: similar comment about maybe using `as` or adding a runtime check

package.json
achou11: for posterity, i believe this is an issue with `prettier-standard`, see https://github.com/sheerun/prettier-standard/issues/111#issuecomment-782924961. probably not worth trying to fix right now since we're only creating declaration files, but once we do a larger scale migration to TS, we should hopefully have this fixed :)

achou11 (approved):
Overall seems good! had a couple of comments that may require some action. This is just the integration right? i.e. none of the queries/mutations are being used in the PR

package.json
```diff
@@ -34,10 +34,11 @@
     "start": "electron . --disable-http-cache",
     "server": "node index.js --headless",
     "dev": "electron . --debug --disable-http-cache",
-    "postinstall": "npm-run-all -p patch-package extract-presets",
+    "postinstall": "npm-run-all -p patch-package extract-presets electron-rebuild",
```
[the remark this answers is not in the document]
ErikSin: ahh, that is also why the CI is breaking, I believe!

src/renderer/hooks/useMapServerQuery.js
```diff
@@ -0,0 +1,15 @@
...
+
+export function useMapServerQuery (resourcePath, enabled) {
```
achou11, line 9: wondering if the idiomatic approach to using react-query is to create a query for each kind of resource. for example instead of the `resourcePath` determining what resource we're working with, it could be:

```js
function useGetStylesQuery(enabled) {
  return useQuery(['styles'], ...)
}

function useGetStyleQuery(styleId, enabled) {
  return useQuery(['styles', styleId], ...)
}

// etc...
```

Again, not super familiar with react query best practices but a quick look at the docs seems to lean towards this approach

src/renderer/hooks/useMapServerMutation.js
```diff
@@ -0,0 +1,26 @@
...
+
+export function useMapServerMutation (mutationType, resourcePath) {
```
achou11, line 8: similar question to https://github.com/digidem/mapeo-desktop/pull/708/files#r1050173134 about creating multiple hooks based on resource

src/renderer/hooks/useMapServerQuery.js
ErikSin: this is a JS file, not TS, so I don't believe this will work

achou11: ah right sorry, forgot about that 😅 in that case, take it or leave the suggestion

ErikSin: I am quite excited to switch our bundler and actually be able to use Ts(x) files....

src/renderer/hooks/useMapServerQuery.js
ErikSin: Yeah, Gregor and I talked a little about this. We were thinking that we did not want to create too many abstractions, given the simplicity of the API. So we decided to just use the one hook, and have the resource path as the keys. I think when we do more complicated things in the future this would be the way to go.

achou11: got it, figured it was discussed but asking for my own education whenever i eventually touch this stuff 😄

src/main/index.js
ErikSin: Im not sure why, but adding `.default` just breaks the app. 
image

src/renderer/hooks/useMapServerQuery.js
ErikSin: yes, you are right, I will update

src/main/index.js
achou11: ah yeah that's on the map server side then, we may have removed that functionality for some reason and didn't update the docs. nevermind then! 😄

src/renderer/hooks/useMapServerQuery.js
ErikSin: updated via: https://github.com/digidem/mapeo-desktop/pull/708/commits/d165cf34175c8e0bb856056c6f0f2c60ccd1a531

package.json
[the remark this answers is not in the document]
ErikSin: updated: https://github.com/digidem/mapeo-desktop/pull/708/commits/d165cf34175c8e0bb856056c6f0f2c60ccd1a531

PR #717, opened 2023-01-24
Es/chore settings about and config

ErikSin:
Added Settings menu, and ability to change config.

### DIFF
--- a/src/renderer/components/Settings/AboutMapeo.js
+++ b/src/renderer/components/Settings/AboutMapeo.js
@@ -7,7 +7,7 @@ import buildConfig from '../../../build-config'
 const m = defineMessages({
   // Used to indicate the version of mapeo the user is using
   version: 'Version',
-  //Title for the About Mapeo Page
+  // Title for the About Mapeo Page
   aboutMapeo: 'About Mapeo',
   // Used to indicate the variant of mapeo the user is using
   variant: 'Variant'
@@ -28,7 +28,7 @@ export const AboutMapeo = () => {
       <Paper className={classes.banner}>
         <Typography variant='h5'>{t(m.aboutMapeo)}</Typography>
       </Paper>
-      <div style={{ padding: 40, backgroundColor: '#f6f6f6', height: '100%' }}>
+      <div style={{ padding: 20, backgroundColor: '#f6f6f6', height: '100%' }}>
         <div className={classes.listItem}>
           <Typography style={{ fontWeight: 500 }}>
             {t(m.version) + ':'} &nbsp;


src/renderer/components/Settings/ProjectConfig.js
```diff
@@ -0,0 +1,130 @@
...
+      properties: ['openFile']
+    })
+    if (result.canceled) return
+    if (!result.filePaths || !result.filePaths.length) return
+
+    ipcRenderer.send('update-config', result.filePaths[0])
+    window.location.reload()
+    //   userConfig.importSettings(result.filePaths[0], err => {
+    //     if (err) return onerror(err)
+    //     ipc.send('reload-config', async err => {
```
ErikSin, line 42: I need to call this IPC call. this IPC call originally was being sent from the main process to [mapeo-core](https://github.com/digidem/mapeo-desktop/blob/71276630bf09497529fb1fa1139261cf6beedba8/src/background/mapeo-core/index.js#L128), which is a background process.

Now we are sending this from a renderer process. Do you know the best way to do this? @gmaclennan 

Should i send an IPC call to the main thread, and then that IPC sends and ipc call to mapeo-core?

gmaclennan: @ErikSin we set up a MessageChannel to communicate between renderer processes. There is currently a message channel established between the [`mapeo-core`](https://github.com/digidem/mapeo-desktop/tree/master/src/background/mapeo-core) background process and the frontend. The message handlers are exported from [`mapeo-core/index.js`](https://github.com/digidem/mapeo-desktop/blob/master/src/background/mapeo-core/index.js), and the coordination between the message channel and the handlers is in [`src/background/index.js`](https://github.com/digidem/mapeo-desktop/blob/master/src/background/index.js). The initial connection of the MessageChannels needs to be coordinated via the main process, which is done here: https://github.com/digidem/mapeo-desktop/blob/master/src/background/index.js. Finally on the front-end renderer process the channel is connected in https://github.com/digidem/mapeo-desktop/blob/master/src/renderer/index-preload.js.

If you want to do IPC with mapeo-core, it is a question of adding extra handler methods to [`mapeo-core/index.js`](https://github.com/digidem/mapeo-desktop/blob/master/src/background/mapeo-core/index.js) and calling them from within https://github.com/digidem/mapeo-desktop/blob/master/src/renderer/new-api.js, which has access to the message channel.

If you want to do IPC to a different background renderer process (e.g. not mapeo-core), then you will need to set up a new channel, and attach it to `window` in the renderer. I can help with more details if that is what you need.

PR #720, opened 2023-02-08
chore: upgrade electron to v22

achou11:
Towards #715 

Changes:

- Upgrade Electron to 22.3.2
  - Requires migration to [`@electron/remote`](https://github.com/electron/remote)
    - Requires changing usage of [`electron-is-dev`](https://github.com/sindresorhus/electron-is-dev) because it no longer works in the renderer process as a result of `remote` removal
  - Upgrade map server to allow usage of better-sqlite3 > 7, which is required for this version of Electron
- Upgrade Node to 16.17.1 for dev environment, to match the version that Electron runs
  - Patches `ecstatic` because of invalid shebang in source file
- Upgrade `react-intl` to 6.2.8 
  - Not sure if this was necessary but based on https://github.com/digidem/mapeo-desktop/issues/715#issuecomment-1402799152, seemed necessary. (not sure how or if that note was related to Electron specifically though)
  - Can potentially revert this since it may be out of scope for the electron upgrade
- Fixes race condition issue with map server initialization when app starts for the first time
- Small removals/cleanups for unused things for development

Questions:

- Is it okay to use npm 8 instead of 6? how does that affect native builds if at all?
- What changes are needed in `build.config.js` to get things working? Seems like we do a lot of manual file picking in there so may need to update based on deps changes
- Should we revert upgrade of `react-intl` if possible? (see above for details)

Potential changes to make (either here or in a follow-up):

- Complete migration away from using `remote` apis.
- Improve usage of files that are used in different processes (e.g. `store.js`, `client-ipc.js`, `logger.js`, etc)

PR #721, opened 2023-02-16
chore: remove Flow from project

achou11:
Removes [Flow](https://flow.org) from the project in its entirety, with the help of [flow-remove-types](https://flow.org/en/docs/tools/flow-remove-types/).

~Note that this PR currently points to `es/chore-electron-upgrade` (see #720). Not entirely necessary but useful to know if these changes work given the upgrade (which seems like a yes). Open to updating so that it can point directly to master branch though.~ EDIT: Going to point this PR to `master` instead so that it's not blocked by the Electron upgrade, since these should be separate


ErikSin (approved):
Given that CI passed, it is running normally on my local environment, and you used the flow cli...I think we can merge this...

Issue #735, opened 2023-07-05
Introduce Zustand Store


ErikSin:
For state management, introduce a Zustand Store. This will be initially used for the Background Map Manager

## To do
- [ ] install Zustand
- [ ] Create Background Map Store (see MapStore type below)
- [ ] Add persistance middleware
- [ ] Persisted state needs to be versioned (for migration purposes)

``` ts
type MapStore = {
    mapStyle:string,
    setMapStyle:(style:string) => void
}

```

PR #738, opened 2023-07-17
Ej/zustand store

lightlii:
### Contributor checklist:

### Description
Adds [zustand](https://docs.pmnd.rs/zustand/getting-started/introduction) state store with persistence using our current [conf](https://www.npmjs.com/package/conf) store. Versioning is achieved with a `STORE_VERSION` constant applied to the store, not sure if this is the best way, open to feedback.

closes https://github.com/digidem/mapeo-desktop/issues/735


lightlii:
ok looks like versioning might be better handled in the conf store? looking at the **migrations** section [here](https://www.npmjs.com/package/conf#migrations)

lightlii:
added another commit with some types on the store. Not sure they're any good though, [Zustand's typescript docs](https://docs.pmnd.rs/zustand/guides/typescript) are a bit vague unfortunately...

lightlii:
Ok I'm slightly more confident in my typings now but still need someone to look over them

lightlii:
lastly, since we now have a means of persisting state through zustand I suggest removing the `createPersistedState`. It would be less confusing to have only a single way of persisting state in the app. There's only a since use of it in the app so its a very minor change and again I can roll it into this PR if you agree @ErikSin ?

ErikSin:
That sounds good! Thanks @lightlii

src/renderer/store.js
ErikSin, line 1: You can use generic's here to get typesafety for all your store slices (For what it is worth....this took me an obnoxiously long time to figure out in JSDOC)

Example:
``` ts

/**
* @typedef {{
*  backgroundMaps: boolean,
*  setBackgroundMapsFlag: (backgroundMaps:boolean) => void,
* }} ExperimentsFlagsStoreSlice
*/

/** 
 * @type {import('zustand').StateCreator<ExperimentsFlagsStoreSlice>}
 */
const experimentsFlagsStoreSlice = (set) => ({
  backgroundMaps: false,
  setBackgroundMapsFlag: backgroundMaps => set({ backgroundMaps }),
})

/**
 * @template T
 * @param {import('zustand').StateCreator<T>} slice
 * @param {string} storeName
 * @returns {import('zustand').UseBoundStore<import('zustand').StoreApi<T>>}
 * 
 */
const createPersistedStore = (slice, storeName) =>
  create(
    persist(slice, {
      name: storeName,
      storage: createJSONStorage(() => storage),
    }),
  )

const useExperimentsFlagsStore = createPersistedStore(experimentsFlagsStoreSlice, 'experiments-flags')

```

Now `useExperimentsFlagsStore` will be typed and you will have typesafety while using it

src/renderer/store.js
```diff
@@ -0,0 +1,90 @@
...
+   * @returns {Promise<void>}
+   */
+  setItem: async (key, state) => {
+    await store.set(key, state)
+  },
+  /**
+   * @param {string}  key - A string key to reference the stored item to be removed.
+   */
+  removeItem: async key => {
+    await store.delete(key)
```
ErikSin, line 28: `set`, `get`, `delete` are synchronous and do not need "await"

src/renderer/store.js
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * @type {import('zustand/middleware').StateStorage}
+ */
+const storage = {
+  /**
+   * @param {string}  key - A string key to reference the stored item by.
+   * @returns {Promise<string | null>}
+   */
+  getItem: async key => {
+    return (await store.get(key)) || null
```
ErikSin, line 14: there is a typing issue here. Typescript is expecting `string` or `null`. `store.get` return `unknown`. You will have to cast a type here

src/renderer/store.js
```diff
@@ -0,0 +1,90 @@
+import { create } from 'zustand'
+import { persist, createJSONStorage } from 'zustand/middleware'
+import store from '../persist-store'
+
+/**
+ * @type {import('zustand/middleware').StateStorage}
+ */
+const storage = {
```
ErikSin, line 8: We should probably set a [schema to storage](https://github.com/sindresorhus/electron-store#schema)

ErikSin (changes requested):
I set up a meeting to figure out your pre-commit steps that don't seem to be working

In order to utilize typescript, you need to add "//@ts-check" to the first line of every file. store.js has some ts errors that you are not seeing currently.

I think `store.js` should be stored in the folder `hooks` since it just returns hooks

src/renderer/store.js
lightlii: Thanks! I also spent a daft amount of time trying to work out how to do it in JSDOC, glad you figured out the proper way!

src/renderer/store.js
lightlii: Not sure what that would look like or how best implement it... We can chat about that later though

src/renderer/components/Home.js
```diff
@@ -15,12 +15,12 @@ import ErrorDialog from './dialogs/Error'
 import ChangeLanguage from './dialogs/ChangeLanguage'
 import TitleBarShim from './TitleBarShim'
 import { defineMessages, useIntl } from 'react-intl'
-import createPersistedState from '../hooks/createPersistedState'
 import SyncView from './SyncView'
 import { STATES as updateStates, UpdaterView, UpdateTab } from './UpdaterView'
 import useUpdater from './UpdaterView/useUpdater'
 import Loading from './Loading'
 import buildConfig from '../../build-config'
+import { usePersistedUiStore } from '../store'
```
ErikSin, line 23: Don't forget to update the import!

Issue #734, opened 2023-07-05
Refactor Map Style Provider


ErikSin:
** Unfinished ticket ** 
In the `Observations` Tab the Map Style is determined by [Map Style Provider](https://github.com/digidem/mapeo-desktop/blob/master/src/renderer/components/MapFilter/MapStyleProvider.js). This is the desired behaviour when the `Background Maps` is set to disabled (in feature flags). When `Background Maps` is enabled, we want the Map Style to be determined by the chosen map from [`useMapServerQuery` hook](https://github.com/digidem/mapeo-desktop/blob/b258b84cab84db0011a9e534cabd1668e7a46eb1/src/renderer/hooks/useMapServerQuery.d.ts#L19). Refactor the `Map Style Provder` to use `useMapServerQuery`.

## To Do:
- [ ] in `MapStyleProvider`, check to see if Background Map Manager `feature flag` is on
- [ ] if feature flag is off, use current logic
- [ ] if feature flag is on,

PR #741, opened 2023-07-27
Ej/ useMapStylesQuery hook

lightlii:
### Description

Adds two new hooks to fetch map styles using new react-query when backgroundmaps enabled, otherwise reverts to legacy api.

closes https://github.com/digidem/mapeo-desktop/issues/734


lightlii:
ok I've basically mirrored how this is working on MM (probably should have looked at that sooner 🙄).

Not 100% sure about the objects beeing returned from the legacy query [here](https://github.com/digidem/mapeo-desktop/blob/837657d9687f8bb17c25a9279c7eeecc2add1ab7/src/renderer/hooks/useMapStylesQuery.js#L47C12-L47C12) and [here](https://github.com/digidem/mapeo-desktop/blob/837657d9687f8bb17c25a9279c7eeecc2add1ab7/src/renderer/hooks/useMapStylesQuery.js#L56), can you take a look?

src/renderer/hooks/useMapStylesQuery.js
```diff
@@ -0,0 +1,69 @@
...
+const ONLINE_STYLE_URL = 'mapbox://styles/mapbox/outdoors-v10'
+
+const m = defineMessages({
+  // The name of the default background map
+  defaultBackgroundMapName: 'Default',
+  // The name of the legacy offline background map
+  offlineBackgroundMapName: 'Offline Map'
+})
+
+export const useMapStylesQuery = () => {
```
ErikSin, line 21: can you enforce the typing to `UseQueryResult<MapServerStyleInfo[]>. Both functions are returning that, but since one is not ecplicitly that, 2 different types are coming up in the ts definitions

ErikSin (changes requested):
Can you rebase/merge. there should only be the one file that is being checked

PR #744, opened 2023-08-14
Ej/background map selection

lightlii:
Currently WIP. PR https://github.com/digidem/mapeo-desktop/pull/741 contains work that needs confirming and merging so this can be completed

image


src/renderer/components/SettingsView/BackgroundMaps.js
```diff
@@ -0,0 +1,85 @@
...
+            <br />
+          </Typography>
+        </div>
+      ) : offlineMap ? (
+        <BackgroundMapInfo
+          key={offlineMap.id}
+          size={offlineMap.bytesStored}
+          id={offlineMap.id}
+          unsetMapValue={unsetMapValue}
+          url={offlineMap.url}
```
ErikSin: Default map should be the available on this list.

Also if this list is longer than the screen, the user is unable to scroll

image

src/renderer/components/MapFilter/MapView/BackgroundMapSelector.js
```diff
@@ -0,0 +1,168 @@
...
+  )
+
+  const defaultMapStyle = useDefaultMapStyle()
+  const { isLoading, data } = useMapStylesQuery()
+
+  const mapStyles =
+    backgroundMapsEnabled && data ? [defaultMapStyle, ...data] : data
+
+  return (
+    <Slide direction='up' in={active} mountOnEnter unmountOnExit>
```
ErikSin: Instead of a bottoms sheet, can we just make this a [dialog](https://mui.com/material-ui/react-dialog/) to follow desktop convention.

src/renderer/components/MapFilter/MapView/BackgroundMapSelector.js
```diff
@@ -0,0 +1,168 @@
...
+                        }}
+                        selected={isSelected}
+                        styleUrl={url}
+                        title={name}
+                      />
+                    </div>
+                  </React.Fragment>
+                )
+              })}
+          </Box>
```
ErikSin: if there are lots of map the user is unable to scroll to see them. When you convert this to a dialog, you may need to add a scroll view, but the user should be able to see the majority of the maps if you make it a full screen dialog

image

src/renderer/hooks/useMapStylesQuery.js
ErikSin, line 1: I think the architecture of this is a little confusing.

We could simplify this by just exporting 1 hook, a react query (server state) hook.

This`react-query-hook` that always returns the list of available styles. This would mean that when background maps is disabled, it just returns an array with 1 map style. And when background maps is enabled, it returns an array (which also includes the default map style). This hook should encapsulate all this logic. If returning the react-query is too complicated, you could just return the list of styles and encapsulate the loading state inside of the hook (aka the hook does not have to return a react-query). It should also handle whether experiments flag is turned on or not.

And then the zustand persisted store, `useBackgroundMapStore`, should handle which map style is used. What that does mean is the map should not be shown while the server state hook is loading. Therefore zustand does not have to deal with any of the loading state, and only deals with returning the persisted map style as chosen by the user. As well the zustand store does not need to deal with  `experimentsFlag`, and is scoped to just serving the map. 

Right now, the hook serving the mapStyle and the hook dealing with server state are both dealing loading state AND the experiments flag, which will make it hard to maintain in the future . As well, because there are so many hooks, the data that is being returned is a combination of the hooks, that again will be hard to maintain in the future.

Essentially there should be 2 exported hooks:
1. zustand state hook that returns the map being used (which is already created)
2. A hook that serves list of avialable maps, that encapsulates `experimentsFlag` and `loading state`

And those can be used in both your modal and in the setting/background map page, without having to transform any data.

ErikSin (changes requested):
I left a comment about architectural changes needeed for the hooks.

lightlii:
I've made a bit of a refactor but I'm not sure it's exactly what you meant. We still need the `useSelectedMapStyle` hook since it needs to run `useMapStylesQuery` and we're not in a react component

src/renderer/hooks/store.js
```diff
@@ -52,15 +56,15 @@ const experimentsFlagsStoreSlice = (set, get) => ({
...
+ *  mapStyle: MapStyle | null,
+ *  setMapStyle: (mapStyle: MapStyle) => void
  * }} BackgroundMapStoreSlice
  */
 /**
  * @type {import('zustand').StateCreator<BackgroundMapStoreSlice>}
  */
 const backgroundMapStoreSlice = (set, get) => ({
-  mapStyle: '',
+  mapStyle: null,
```
ErikSin: instead of null, can you make it just default to the `defaultMap`

src/renderer/hooks/useMapStylesQuery.js
```diff
@@ -18,29 +18,49 @@ const m = defineMessages({
   offlineBackgroundMapName: 'Offline Map'
 })
 
-export const useMapStylesQuery = () => {
+export const useMapStylesQuery = (enabled = true) => {
```
ErikSin, line 30: I dont think I understand what the difference between `enable` and `backgroundMapsEnabled`

src/renderer/components/SettingsView/BackgroundMaps.js
```diff
@@ -0,0 +1,84 @@
...
+          <Typography variant='h4'> {t(m.mapBackgroundTitle)}</Typography>
+
+          <Typography variant='body1'>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+            eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim
+            ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+            aliquip ex ea commodo consequat. Duis aute irure dolor in
+            reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
+            pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
+            culpa qui officia deserunt mollit anim id est laborum.
```
ErikSin: replace with (as a translatable object): "The Background Map in Mapeo is displayed on the observation screen and is used as a background for the observations you collect. This new pilot feature allows you to add your own custom maps and switch between multiple maps. Background Maps is currently an advanced feature - an existing map file in .mbtiles format is required for testing."

src/renderer/components/MapFilter/MapView/MapPreviewCard.js
```diff
@@ -0,0 +1,75 @@
...
+import ReactMapboxGl from 'react-mapbox-gl'
+
+import { MAPBOX_ACCESS_TOKEN } from '../../../../../config'
+
+export const MapboxPreview = ReactMapboxGl({
+  accessToken: MAPBOX_ACCESS_TOKEN,
+  dragRotate: false,
+  pitchWithRotate: false,
+  attributionControl: false,
+  injectCSS: false
```
ErikSin, line 10: add: `interactive:false`

src/renderer/components/BackgroundMaps/BackgroundMapInfo.js
```diff
@@ -0,0 +1,198 @@
...
+        <div>
+          {!isDefault && (
+            <Button variant='outlined' onClick={() => deleteMap()}>
+              <DeleteIcon className={classes.icon} />
+              <Typography style={{ textTransform: 'none' }} variant='subtitle2'>
+                {t(m.deleteStyle)}
+              </Typography>
+            </Button>
+          )}
+        </div>
```
ErikSin, line 159: this is currently not working, but I believe this is a map server thing, so just remove it for now (and the user will not be able to delete a map)

src/renderer/hooks/useMapStylesQuery.js
lightlii: `enable` enables the query. It allows us to disable the querying and manually call `refetch` when we want the data. I can't find the original ticket but it was one of the first thing you asked me to add on this feature!

src/renderer/components/BackgroundMaps/BackgroundMapInfo.js
lightlii: this is working fine for me, lets check together later whats going on

lightlii: (I added a little check so it will set the the default map if you delete the current selected one)

PR #746, opened 2023-08-17
Ej/fix persisted store

lightlii:
Quick fix, reinstates async await syntax when getting an item from the store.

I've also made the config store consistent since zustand was stringifying the store upon save wheras by default the lib stores it as json which was resulting in configs that look like this:

```
{
	"experiments-flags": "{\"state\":{\"backgroundMaps\":false},\"version\":0}",
	"background-maps": "{\"state\":{\"mapStyle\":\"\"},\"version\":0}",
	"ui": "{\"state\":{\"tabIndex\":0},\"version\":0}",
	"state": {
		"maximize": false,
		"fullscreen": false,
		"defaultWidth": 1000,
		"defaultHeight": 800,
		"isMaximized": false,
		"show": false,
		"x": 2199,
		// etc.
	}
}

### CODE REVIEW

File: src/renderer/hooks/store.js
Comment (ErikSin) [inline src/renderer/hooks/store.js:1]: based on the typescript typings, it looks like get item has to return a string, and I assume `createJSONStorage` then just turns it back to an object...its seems like it is working.

playing around with it, I also realize that `getItem`, `setItem`, and `removeItem` don't need jsdoc typings, those are encapasulated it typing of the 'storage'

**Review by ErikSin — APPROVED** [review]
Can you get rid of the unnecessary typing before you merge please and thank you

File: src/renderer/hooks/store.js [hunk 1]
```diff
@@ -12,15 +12,20 @@ const storage = {
...
    * @param {string} state - The state to be persisted - stringified JSON.
    * @returns {void}
    */
   setItem: (key, state) => {
-    store.set(key, JSON.parse(state))
+    if (typeof state === 'string') {
+      store.set(key, JSON.parse(state))
+    } else {
+      store.set(key, state)
+    }
```
ErikSin: with some more testing, it seems like zustand will automatically create your object into a string (so you don't need to do this typecheck)

ErikSin:
You can just get rid of the typings for `getItem`, `setItem`, and `removeItem`, and then you can merge

PR #747, opened 2023-08-30
Es/example of persisted state

ErikSin:
this is incomplete, but I wanted to show an example of encapsulating the logic of whether background maps is on or off directly in the hook.

Also example of doing undefined check inside the useMapStylesQuery hook

### DIFF
--- a/src/renderer/hooks/store.js
+++ b/src/renderer/hooks/store.js
@@ -1,25 +1,18 @@
 // @ts-check
 import { create } from 'zustand'
 import { persist, createJSONStorage } from 'zustand/middleware'
-import { defineMessages } from 'react-intl'
 import store from '../../persist-store'
-
-const m = defineMessages({
-  // The name of the default background map
-  defaultBackgroundMapName: 'Default'
-})
-
-const DEFAULT_MAP = {
-  id: '487x2pc8ws801avhs5hw58qnxc',
-  url: 'mapbox://styles/mapbox/outdoors-v10',
-  bytesStored: 0,
-  name: m.defaultBackgroundMapName,
-  isImporting: false,
-  isDefault: true
-}
+import { DEFAULT_MAP } from './useMapStylesQuery'
 
 /**
- * @typedef { {id: string, url: string,bytesStored: number,name: import('react-intl').MessageDescriptor}} MapStyle
+ * @typedef { {
+ * id: string,
+ * url: string,
+ * bytesStored: number,
+ * name: import('react-intl').MessageDescriptor | string,
+ * isDefault?: boolean,
+ * isImporting?: boolean
+ * }} MapStyle
  */
 
 /**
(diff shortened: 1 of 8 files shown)


src/renderer/hooks/store.js
```diff
@@ -86,11 +107,26 @@ export const useExperimentsFlagsStore = createPersistedStore(
...
+const useBackgroundMapState = createPersistedStore(
   backgroundMapStoreSlice,
   'background-maps'
 )
 export const usePersistedUiStore = createPersistedStore(
   persistedUiStoreSlice,
   'ui'
 )
+
+export const useBackgroundMapStore = () => {
```
ErikSin, line 117: Now this hooks encapsulates the logic for whether background maps is enable or not. And can only show and set the maps that are relevant to that state. No need to the developer to do a check everytime.

src/renderer/hooks/store.js
```diff
@@ -86,11 +107,26 @@ export const useExperimentsFlagsStore = createPersistedStore(
...
+    store => store.backgroundMaps
+  )
+
+  const [mapStyle, setMapStyle] = useBackgroundMapState(store =>
+    backgroundMapsEnabled
+      ? [store.mapStyleMapServer, store.setMapStyleServer]
+      : [store.mapStyleLegacy, store.setMapStyleLegacy]
+  )
+
+  return { mapStyle, setMapStyle }
```
ErikSin: ideally this would return an array, but I am not too sure how to the equivalent of `as const`

src/renderer/hooks/useMapStylesQuery.js
```diff
@@ -42,7 +42,9 @@ export const useMapStylesQuery = (enabled = true) => {
         refetch: mapStylesQueryResult.refetch
       }
     : {
-        data: legacyStyleQueryResult.data,
+        data: !legacyStyleQueryResult.data
+          ? [defaultMapStyle]
+          : legacyStyleQueryResult.data,
```
ErikSin: this encapsulates the undefined check so the consuming components do not have to do that

src/renderer/hooks/store.js
```diff
@@ -1,10 +1,25 @@
...
+})
+
+const DEFAULT_MAP = {
+  id: '487x2pc8ws801avhs5hw58qnxc',
+  url: 'mapbox://styles/mapbox/outdoors-v10',
+  bytesStored: 0,
+  name: m.defaultBackgroundMapName,
+  isImporting: false,
+  isDefault: true
+}
```
ErikSin: This is in 2 places, so ideally should just be moved to one place!

src/renderer/hooks/store.js
ErikSin: I figured out how to return an array
image

lightlii:
Hey this makes sense. I've wrapped up the loose ends on this so it can be merged if you're happy with it

PR #759, opened 2025-03-04
chore(ci): get CI working again

achou11:
Updates workflow to get CI running again

Note that this is pointing to `release/v5.6.4` and not `master`.

Summary of changes:

- Updates versions of actions used in the workflow
- Updates the builder config to disable code signing (both macOS and Windows)
  - Not ideal but our certificates for both platforms are expired. Happy to re-enable if we have the means of updating the certificates.
- Updates the version of `electron-builder` to a slightly more recent version in order to incorporate fixes for issues caused by changes in our workflow images.


.github/workflows/nodejs.yml
```diff
@@ -15,10 +15,10 @@ jobs:
     strategy:
       matrix:
         config:
-          - { os: macos-11, arch: x64 }
+          - { os: macos-13, arch: x64 }
           - { os: ubuntu-latest, arch: x64 }
```
achou11: might be worth pinning this. seems like Ubuntu 24 is working so maybe it's fine to set to that?

.github/workflows/nodejs.yml
```diff
@@ -15,10 +15,10 @@ jobs:
...
-          - { os: windows-latest, arch: x64 }
-          - { os: windows-latest, arch: ia32 }
+          - { os: windows-2019, arch: x64 }
+          - { os: windows-2019, arch: ia32 }
```
achou11, line 21: wasn't really sure what version to set this to be seemed like the [master branch](https://github.com/digidem/mapeo-desktop/blob/6769cd520e9907ac1b573074a6b8c1ed70bef89c/.github/workflows/nodejs.yml#L19-L20) updated this to 2019 as well, so went with that

package.json
```diff
@@ -220,7 +220,7 @@
     "dotenv": "^16.4.5",
     "dotenv-webpack": "^8.1.0",
     "electron": "^9.4.4",
-    "electron-builder": "^22.14.10",
+    "electron-builder": "23.6.0",
```
achou11, line 224: potentially dangerous but equally helpful. the main reason for updating this is to fix issues related to running on newer machines. see 119ebdc3518bf50b561fb9aad65bc7334f56e7ce for more details.

read through the changelogs between our previous version and this version and nothing stood out to me as a major issue waiting to happen. however, open to addressing the original issue differently if we're worried about that

.github/workflows/nodejs.yml
achou11: done via https://github.com/digidem/mapeo-desktop/pull/759/commits/1e6ec88f57333690ad21efcc17e1817c0c9ee01a

package.json
gmaclennan: Seems to work so I'm ok with this.

.github/workflows/nodejs.yml
gmaclennan: This change was probably to pin the version, since `windows-latest` became `windows-2022` and now `windows-2025`. Since this works, let's leave this as `windows-2019` for now, however note that this will likely be deprecated soon, since apparently [Github supports the latest 2 LTS versions](https://github.com/actions/runner-images/discussions/5343) and [Windows 2025 is now released](https://www.microsoft.com/en-us/windows-server/blog/2024/11/04/windows-server-2025-now-generally-available-with-advanced-security-improved-performance-and-cloud-agility/) (previous LTS were Windows 2022 and 2019).

gmaclennan:
I added back MacOS code signing, since I have recently re-activated our Apple Developer Account. I also updated notarization to work.

PR #760, opened 2025-03-05
chore: merge changes from 5.6.x releases back into default branch

achou11:
Uses a "merge" approach i.e.

```sh
git checkout -b <new-branch-name>
git merge release/v5.6.4
```

See for list of relevant changes: https://github.com/digidem/mapeo-desktop/compare/master...release/v5.6.4

Opening up a separate identical PR that uses a cherry-picking approach. Not sure which one's preferable or if it really matters.

PR #761, opened 2025-03-05
chore: merge changes from 5.6.x releases back into default branch

achou11:
Functionally similar to #760 but uses a cherry-picking approach i.e. `git cherry-pick ...` for every commit listed in https://github.com/digidem/mapeo-desktop/compare/master...release/v5.6.4


achou11:
discussed elsewhere, but closing since not a priority. can revisit if needed

